PR #1, opened 2020-10-30
Api proposal

maasencioh:
@lpatiny @targos @stropitek what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,23 +7,40 @@ Render 2dplot partition in react.
 
 ## [Storybook](https://zakodium.github.io/react-plot/)
 
-```ts
-interface DataPoints {
-  x: Array<number>;
-  y: Array<number>;
-  label?: string;
-  color?: string;
-}
-
-interface DataAxis {
-  label: string;
-  units?: string;
-  displayGrid?: boolean;
-  logarithmic?: boolean;
-}
-
-const Example = (data: DataPoints, axis: DataAxis) => (
-  <Plot data={data} axis={axis} />
+```tsx
+const Example = () => (
+  <Plot
+    width={500}
+    height={500}
+    colorScheme={d3.schemePaired}
+    margin={{ right: 200 }}
+  >
+    <PlotHeading
+      title="Electrical characterization"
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,21 +1,58 @@
...
+
+```tsx
+const Example = () => (
+  <Plot
+    width={500}
+    height={500}
+    colorScheme={d3.schemePaired}
+    margin={{ right: 200 }}
+  >
+    <PlotHeading
```
lpatiny, line 18: I would rename to Heading and remove the property 'top'. If it is a heading it should be on the top.

README.md
```diff
@@ -1,21 +1,58 @@
...
+      title="Electrical characterization"
+      subtitle="current vs voltage"
+      titleStyle={}
+      substitleStyle={}
+      titleClass=""
+      substitleClass=""
+      position="top"
+    />
+    <PlotLineSeries
```
lpatiny, line 27: Could be renamed to LineSeries

README.md
```diff
@@ -1,21 +1,58 @@
...
+      label="Vg = 7V"
+    />
+    <PlotLineSeries
+      data={{ x: [0, 1, 2, 3, 4, 5], y: [0, 2, 4, 6, 6, 6] }}
+      lineStyle={{ stroke: 'blue', strokeWidth: 2 }}
+      markerStyle={}
+      displayMarker={true}
+      label="Vg = 3V"
+    />
+    <PlotXAxis label="Drain voltage [V]" showTicks={true} tickStep={1} />
```
lpatiny, line 40: could also be renamed to XAxis, YAxis, Legend

README.md
maasencioh: If I have several plots I could want to have this on the bottom, and I like the `ComponentSubcomponent` style for subcomponents, in order to make it easy to understand when you imported (Header could be your layout header, menu header, etc)

maasencioh:
Another option will be to use `Plot.Heading` or `Plot.XAxis` for example, but I prefer clear names

targos (approved):

PR #2, opened 2020-11-11
Base implementation

maasencioh:
Not a big fan of long PR, so the idea is just to display the line series

https://deploy-preview-2--amazing-kilby-d9c1d7.netlify.app/?path=/story/example-plot--control
Screenshot 2020-11-12 105739

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ const Example = () => (
     colorScheme={d3.schemePaired}
     margin={{ right: 200 }}
   >
-    <PlotHeading
+    <Plot.Heading
       title="Electrical characterization"
       subtitle="current vs voltage"
       titleStyle={}
@@ -24,22 +24,22 @@ const Example = () => (
       substitleClass=""
       position="top"
     />
-    <PlotLineSeries
+    <Plot.LineSeries
       data={{ x: [0, 1, 2, 3, 4, 5], y: [0, 1, 2, 3, 3, 3] }}
       lineStyle={{ stroke: 'blue', strokeWidth: 2 }}
       curve={d3.curveCardinal(0.5)}
       label="Vg = 7V"
     />
-    <PlotLineSeries
+    <Plot.LineSeries
       data={{ x: [0, 1, 2, 3, 4, 5], y: [0, 2, 4, 6, 6, 6] }}
       lineStyle={{ stroke: 'blue', strokeWidth: 2 }}
       markerStyle={}
       displayMarker={true}
       label="Vg = 3V"
     />
-    <PlotXAxis label="Drain voltage [V]" showTicks={true} tickStep={1} />
-    <PlotYAxis label="Drain current [mA]" showTicks={true} tickStep={1} />
-    <PlotLegend position="right" direction="down" />
+    <Plot.XAxis label="Drain voltage [V]" showTicks={true} tickStep={1} />
+    <Plot.YAxis label="Drain current [mA]" showTicks={true} tickStep={1} />
+    <Plot.Legend position="right" direction="down" />
   </Plot>
 );
 ```
--- a/README.md
+++ b/README.md
@@ -25,13 +25,27 @@ const Example = () => (
       position="top"
     />
     <Plot.LineSeries
-      data={{ x: [0, 1, 2, 3, 4, 5], y: [0, 1, 2, 3, 3, 3] }}
+      data={[
+        { x: 0, y: 0 },
+        { x: 1, y: 1 },
+        { x: 2, y: 2 },
+        { x: 3, y: 3 },
+        { x: 4, y: 3 },
+        { x: 5, y: 3 },
+      ]}
       lineStyle={{ stroke: 'blue', strokeWidth: 2 }}
       curve={d3.curveCardinal(0.5)}
       label="Vg = 7V"
     />
     <Plot.LineSeries
-      data={{ x: [0, 1, 2, 3, 4, 5], y: [0, 2, 4, 6, 6, 6] }}
+      data={[
+        { x: 0, y: 0 },
+        { x: 1, y: 2 },
+        { x: 2, y: 4 },
+        { x: 3, y: 6 },
+        { x: 4, y: 6 },
+        { x: 5, y: 6 },
+      ]}
       lineStyle={{ stroke: 'blue', strokeWidth: 2 }}
       markerStyle={}
       displayMarker={true}
--- a/README.md
+++ b/README.md
@@ -25,27 +25,13 @@ const Example = () => (
       position="top"
     />
     <Plot.LineSeries
-      data={[
-        { x: 0, y: 0 },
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -59,5 +59,12 @@
     "react-is": "^16.13.1",
     "typescript": "^4.0.3"
   },
-  "dependencies": {}
+  "dependencies": {
+    "@types/d3-array": "^2.5.1",
+    "@types/d3-scale": "^3.2.1",
+    "@types/d3-shape": "^2.0.0",
```
targos, line 65: devDependencies

src/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export function getMin(actionMin: number, stateMin: number | null | undefined) {
+  return stateMin === null || stateMin === undefined || stateMin >= actionMin
+    ? actionMin
+    : stateMin;
+}
+
+export function splitChildren(children: React.ReactNode[]) {
+  return React.Children.toArray(children).reduce<PlotChildren>(
```
targos, line 19: The implementation of this function would be much simpler with a for..of loop or by using array.forEach

src/index.ts
```diff
@@ -1 +1,4 @@
-export * from './Plot';
+import Plot from './Plot';
+import LineSeries from './components/LineSeries';
+
+export { Plot as default, LineSeries };
```
targos, line 4: Can we export everything with a name? It's not useful anyway to only import the default, so `import { Plot, LineSeries } from 'react-plot'` is simpler than `import Plot, { LineSeries } from 'react-plot'`

src/Plot.tsx
```diff
@@ -1,3 +1,47 @@
...
+  const { invalidChild, lineSeries } = splitChildren(children);
+  if (invalidChild) {
+    // eslint-disable-next-line no-console
+    console.error('Only compound components of Plot are displayed');
+  }
+
+  // Propagate props to children
+  const lineSeriesInjected = lineSeries.map((series) =>
+    React.cloneElement(series, { width, height }),
+  );
```
targos, line 37:
```suggestion
```

src/components/LineSeries.tsx
```diff
@@ -0,0 +1,48 @@
...
+  lineStyle,
+  width,
+  height,
+}: LineSeriesRenderProps) {
+  // Get scales from context
+  const { xMin, xMax, yMin, yMax } = usePlotContext();
+  if ([xMin, xMax, yMin, yMax].includes(undefined)) return null;
+
+  const xScale = scaleLinear().range([0, width]).domain([xMin, xMax]);
+  const yScale = scaleLinear().range([height, 0]).domain([yMin, yMax]);
```
targos, line 41: pleas

targos, line 41: You can create the scales in Plot and pass them to the context. That way you won't need to pass width and height to this component.

src/Plot.tsx
```diff
@@ -1,3 +1,47 @@
...
+  switch (action.type) {
+    case 'newData': {
+      return {
+        ...state,
+        xMin: getMin(action.value.xMin, state.xMin),
+        xMax: getMax(action.value.xMax, state.xMax),
+        yMin: getMin(action.value.yMin, state.yMin),
+        yMax: getMax(action.value.yMax, state.yMax),
+        labels: [...state.labels, action.value.label],
+      };
```
targos, line 17: This implementation is problematic because if a series is removed, the boundaries are not updated.
We have to keep the data from all series separated in the state. I think we should generate an id for the serie in the Serie component and use that to identify it in the main state (using a Map or a Record).

src/Plot.tsx
```diff
@@ -1,3 +1,47 @@
...
+  }
+}
+
+export default function Plot({ width, height, children }: PlotProps) {
+  const [state, dispatch] = useReducer(reducer, { labels: [] }, undefined);
+
+  const { invalidChild, lineSeries } = splitChildren(children);
+  if (invalidChild) {
+    // eslint-disable-next-line no-console
+    console.error('Only compound components of Plot are displayed');
```
targos, line 31: let's just throw an error here so the user knows immediately if they do something wrong

src/components/LineSeries.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+interface LineSeriesRenderProps {
+  width?: number;
+  height?: number;
+  data: Series[];
+  lineStyle?: CSSProperties;
+}
+
+export default function LineSeries(props: LineSeriesProps) {
+  const { data, label, ...otherProps } = props;
```
targos, line 17: here you can add a `const [id] = useState(generateId)`

(generateId can just be a function that creates an id with an incremented number like `series-${++number}`)

README.md
```diff
@@ -24,22 +24,36 @@ const Example = () => (
       substitleClass=""
       position="top"
     />
-    <PlotLineSeries
-      data={{ x: [0, 1, 2, 3, 4, 5], y: [0, 1, 2, 3, 3, 3] }}
```
targos, line 28: we would like to keep this way of defining the data. I'm sure it's possible to work with this in d3

src/Plot.tsx
maasencioh: why?

targos: explained in the next comment

README.md
maasencioh: It's possible, but is not meant to be used like that, so will be more expensive to internally change the format than use https://github.com/mljs/xy-convert in the parent

src/utils.ts
```diff
@@ -1,8 +1,17 @@
+import crypto from 'crypto';
```
targos, line 1: You don't need a cryptographic rng for that. This is a big dependency for a front-end project. Do you see any issue with an auto-increment id?

maasencioh: How will you know the current value?

targos:
```js
let currentValue = 1;

function getNextId() {
  return ++currentValue;
}
```

maasencioh: I send it through the context and keep it in the state, shouldn't be lost in your proposal?

targos: You misunderstand me, I think. I'm just suggesting to change `export function randomString() [}` with `export function getNextId() {}`, nothing else.

src/components/LineSeries.tsx
```diff
@@ -0,0 +1,61 @@
...
+import type { LineSeriesProps, Series } from '../types';
+import { getNextId } from '../utils';
+
+interface LineSeriesRenderProps {
+  data: Series;
+  lineStyle?: CSSProperties;
+}
+
+export default function LineSeries(props: LineSeriesProps) {
+  const [id] = useState(`series-${getNextId()}`);
```
targos, line 15:
```suggestion
  const [id] = useState(() => `series-${getNextId()}`);
```

To avoid generating new ids on every render

PR #4, opened 2020-11-17
Axis

maasencioh:
:eyes: [Preview](https://deploy-preview-4--amazing-kilby-d9c1d7.netlify.app/?path=/story/example-plot--control)

Display axis for plot
- add label
- option to display grid by axis

### DIFF
--- a/src/components/XAxis.tsx
+++ b/src/components/XAxis.tsx
@@ -10,7 +10,7 @@ const XAxis = ({ label, fontSize = 16, showTicks }: AxisProps) => {
   const { xScale, margin, height, width } = usePlotContext();
 
   useLayoutEffect(() => {
-    if (axisRef?.current) {
+    if (axisRef?.current && xScale) {
       const axis = axisBottom(xScale);
       if (showTicks) {
         axis.tickSizeInner((margin?.bottom || 0) + (margin?.top || 0) - height);
--- a/src/components/XAxis.tsx
+++ b/src/components/XAxis.tsx
@@ -5,14 +5,14 @@ import React, { useLayoutEffect, useRef } from 'react';
 import { usePlotContext } from '../hooks';
 import type { AxisProps } from '../types';
 
-const XAxis = ({ label, fontSize = 16, showTicks }: AxisProps) => {
+const XAxis = ({ label, fontSize = 16, showGridLines }: AxisProps) => {
   const axisRef = useRef(null);
   const { xScale, margin, height, width } = usePlotContext();
 
   useLayoutEffect(() => {
     if (axisRef?.current && xScale) {
       const axis = axisBottom(xScale);
-      if (showTicks) {
+      if (showGridLines) {
         axis.tickSizeInner((margin?.bottom || 0) + (margin?.top || 0) - height);
       }
 
@@ -20,11 +20,11 @@ const XAxis = ({ label, fontSize = 16, showTicks }: AxisProps) => {
         .call(axis)
         .call((g) => {
           g.selectAll('.tick line')
-            .attr('stroke-opacity', showTicks ? 0.5 : 1)
-            .attr('stroke-dasharray', showTicks ? '2,2' : '0');
+            .attr('stroke-opacity', showGridLines ? 0.5 : 1)
+            .attr('stroke-dasharray', showGridLines ? '2,2' : '0');
           g.selectAll('.tick text').attr(
             'transform',
-            `translate(0,${showTicks ? 6 : 0})`,
+            `translate(0,${showGridLines ? 6 : 0})`,
           );
         });
     }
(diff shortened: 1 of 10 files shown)


src/components/XAxis.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { select } from 'd3-selection';
+import React, { useRef } from 'react';
+
+import { usePlotContext } from '../hooks';
+import type { AxisProps } from '../types';
+
+const XAxis = ({ label, fontSize = 16, showTicks }: AxisProps) => {
+  const axisRef = useRef(null);
+  const { xScale, margin, height, width } = usePlotContext();
+  if (axisRef?.current) {
```
targos, line 11: This is wrong, you should never have side-effects directly in the render. this code should be in a `useEffect`

src/components/YAxis.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { select } from 'd3-selection';
+import React, { useRef } from 'react';
+
+import { usePlotContext } from '../hooks';
+import type { AxisProps } from '../types';
+
+const YAxis = ({ label, fontSize = 16, showTicks }: AxisProps) => {
+  const axisRef = useRef(null);
+  const { yScale, margin, height, width } = usePlotContext();
+  if (axisRef?.current) {
```
targos, line 11: ditto

src/components/XAxis.tsx
stropitek: it's even better to do it in a `useLayoutEffect` because the effect will be fired synchronously after the dom was mutated by react, and this way it will save the browser one paint, which is an intensive task

maasencioh: Thank you guys

targos:
the story fails for me: 

image

maasencioh:
@targos I just realized in other branch, sorry, is fixed already

src/components/YAxis.tsx
```diff
@@ -0,0 +1,57 @@
...
+import React, { useLayoutEffect, useRef } from 'react';
+
+import { usePlotContext } from '../hooks';
+import type { AxisProps } from '../types';
+
+const YAxis = ({ label, fontSize = 16, showTicks }: AxisProps) => {
+  const axisRef = useRef(null);
+  const { yScale, margin, height, width } = usePlotContext();
+
+  useLayoutEffect(() => {
```
targos, line 12: I discussed with Daniel and we think we should follow React's official recommendations and `useEffect` instead.

src/components/XAxis.tsx
```diff
@@ -0,0 +1,60 @@
...
+import React, { useLayoutEffect, useRef } from 'react';
+
+import { usePlotContext } from '../hooks';
+import type { AxisProps } from '../types';
+
+const XAxis = ({ label, fontSize = 16, showTicks }: AxisProps) => {
+  const axisRef = useRef(null);
+  const { xScale, margin, height, width } = usePlotContext();
+
+  useLayoutEffect(() => {
```
targos, line 12: ditto

src/components/XAxis.tsx
```diff
@@ -0,0 +1,60 @@
...
+          g.selectAll('.tick line')
+            .attr('stroke-opacity', showTicks ? 0.5 : 1)
+            .attr('stroke-dasharray', showTicks ? '2,2' : '0');
+          g.selectAll('.tick text').attr(
+            'transform',
+            `translate(0,${showTicks ? 6 : 0})`,
+          );
+        });
+    }
+  });
```
targos, line 31: I suggest to add an array of dependencies with the values that are captured by the callback. Doesn't the linter complain about it missing? Anyway, to avoid running this code on every render, we need the array.

src/components/YAxis.tsx
```diff
@@ -0,0 +1,57 @@
...
+          g.selectAll('.tick line')
+            .attr('stroke-opacity', showTicks ? 0.5 : 1)
+            .attr('stroke-dasharray', showTicks ? '2,2' : '0');
+          g.selectAll('.tick text').attr(
+            'transform',
+            `translate(${showTicks ? -6 : 0},0)`,
+          );
+        });
+    }
+  });
```
targos, line 30: ditto

src/types.ts
```diff
@@ -35,4 +38,19 @@ export type ReducerActions =
 export interface PlotChildren {
   invalidChild: boolean;
   lineSeries: ReactElement<LineSeriesProps>[];
+  axis: Record<'x' | 'y', ReactElement<AxisProps> | null>;
+}
+
+export interface AxisProps {
+  fontSize?: number;
+  label?: string;
+  showTicks?: boolean;
```
targos, line 47: can we call it `showGridLines` instead ? the ticks are the little bars coming out of the axes

maasencioh: I was also thinking about the name, I followed the D3 suggestion, but I prefer your idea

src/components/YAxis.tsx
maasencioh: Just curious, why?

targos: https://reactjs.org/docs/hooks-reference.html#uselayouteffect

targos (approved):

PR #9, opened 2020-11-20
feat: add markers to line

maasencioh:


src/components/LineSeries.tsx
```diff
@@ -54,16 +61,36 @@ function LineSeriesRender({ data, lineStyle, label }: LineSeriesRenderProps) {
...
+    const markers = !displayMarker
+      ? null
+      : series.map(({ x, y }, i) => (
+          <circle
+            // eslint-disable-next-line react/no-array-index-key
+            key={`markers-${i}`}
+            cx={x}
+            cy={y}
+            r="3px"
+            fill={color}
```
lpatiny, line 75: this should be a 'style' property so that everything can be customised using any of the SVG style property. By default it could be style: {fill: color}

lpatiny (commented):
Need also to think how to deal with scatter plots. Actually a scatter plot is a lineSeries without the line and only the markers.

targos:
A scatter plot is not a line series without the line because a line series cannot go back on the X axis.

src/components/LineSeries.tsx
maasencioh: There's any reason to change only the markers colors?

PR #14, opened 2020-12-01
feat: add label format

maasencioh:
Compatible with [d3-format](https://github.com/d3/d3-format)


targos:
Why not call it `tickFormat` if it's for the ticks?

PR #27, opened 2020-12-08
Axis boundaries

maasencioh:


targos:
The updates don't quite work. Try this in the story to increase xMax one by one using the controls. The blue graph starts disappearing progressively

targos:
image

The `clipPath` should probably not be at the level of the series, but instead in a `<g>` that wraps all the contents of the graph

targos (approved):

PR #28, opened 2020-12-08
Scientific notation

maasencioh:


lpatiny:
LGTM

stropitek:
You could divide instead of multiplying on one axis to show the behavior for small numbers

maasencioh:
Thanks @stropitek, good advice

PR #30, opened 2020-12-09
add "paddingXXX" prop to axes

maasencioh:


src/components/YAxis.tsx
```diff
@@ -13,7 +13,9 @@ const YAxis = ({
   labelStyle,
   min,
   max,
-}: AxisProps) => {
+  paddingBottom,
```
lpatiny, line 16: Please define at this level the default value (0) so that you don't need to check after

maasencioh: This value is used in the `Plot.tsx` component, that is updated asynchronously, for this reason, is better, in any case, to check it there

targos: the question is not about checking it, but about having `0` as the default instead of `undefined`

PR #38, opened 2020-12-14
Add example with PCA and multiple plots

targos:
- fix(types): allow Plot to have a single child
- docs: add PCA example with iris dataset

https://deploy-preview-38--amazing-kilby-d9c1d7.netlify.app/?path=/story/examples-iris-dataset--pca

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,8 @@
     "@types/react": "^16.9.52",
     "@zakodium/eslint-config": "^2.1.1",
     "eslint": "^7.11.0",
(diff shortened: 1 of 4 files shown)


targos:
This is an example where https://github.com/zakodium/react-plot/issues/34 will be useful, because we don't want to show both axes in every plot.

targos:
I'm also kind of hacking the use of a LineSeries to do a scatter plot

maasencioh (approved):

PR #42, opened 2020-12-14
General fixes

maasencioh:


src/Plot.tsx
```diff
@@ -137,12 +137,13 @@ export default function Plot({
   }, [state, bottom, top, height]);
 
   const labels = state.series.map(({ label }) => label);
+  const ids = state.series.map(({ id }) => id);
```
targos, line 140: it's better to put this inside the useMemo, otherwise the `ids` array will always be a new array if it's created here

package.json
```diff
@@ -55,6 +55,7 @@
     "@types/d3-array": "^2.5.1",
     "@types/d3-axis": "^2.0.0",
     "@types/d3-scale": "^3.2.1",
+    "@types/d3-scale-chromatic": "^2.0.0",
```
targos, line 58: this is not used here but if you plan to use it later we can keep it

maasencioh: this is used in `Plot.tsx:3`, I don't know why I didn't add it before

targos (approved):

PR #43, opened 2020-12-15
Tick style

maasencioh:
- add tickStyle to the axis, allowing to add styles (uses kebab-case)
- add the source to the stories
- fix the default color palette error

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "typescript": "^4.0.3"
   },
   "dependencies": {
+    "@storybook/addon-storysource": "^6.1.11",
     "csstype": "^3.0.5",
     "d3-array": "^2.8.0",
     "d3-axis": "^2.0.0",
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
   "devDependencies": {
     "@storybook/addon-essentials": "^6.1.10",
     "@storybook/addon-links": "^6.1.10",
+    "@storybook/addon-storysource": "^6.1.11",
     "@storybook/react": "^6.1.10",
     "@types/d3-array": "^2.5.1",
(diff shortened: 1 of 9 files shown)


src/types.ts
```diff
@@ -36,6 +37,7 @@ interface AxisProps {
   label?: string;
   showGridLines?: boolean;
   labelStyle?: CSSProperties;
+  tickStyle?: PropertiesHyphen;
```
targos, line 40: I don't like this because it is different from the other style properties. I'm sure we can avoid it.

package.json
```diff
@@ -69,6 +69,8 @@
     "typescript": "^4.0.3"
   },
   "dependencies": {
+    "@storybook/addon-storysource": "^6.1.11",
```
targos, line 72: devDependencies

targos (approved):

src/utils.ts
```diff
@@ -45,3 +45,7 @@ export function splitChildren(children: React.ReactNode): PlotChildren {
   }
   return { invalidChild, lineSeries, axis, heading, legend };
 }
+
+export function camelToKebab(string: string) {
+  return string.replace(/([a-z0-9]|(?=[A-Z]))([A-Z])/g, '$1-$2').toLowerCase();
```
targos, line 50:
```suggestion
  // eslint-disable-next-line ...
  return string.replace(/([a-z0-9]|(?=[A-Z]))([A-Z])/g, '$1-$2').toLowerCase();
```

maasencioh: I wasn't able to find how to do it with replaces, but here's the link if someone else wants it https://2ality.com/2017/05/regexp-named-capture-groups.html

PR #45, opened 2020-12-16
feat: add scatter plot option

maasencioh:
- [x] add scatter plot
- [ ] add group colors


targos:
there are conflicts

targos:
Could we have the same color for the scatter plot and the regression lines ?

image

maasencioh:
I was planning to create that option in another PR, is the color group option

targos (approved):

Issue #60, opened 2020-12-17
Replace XAxis/YAxis with Axis


targos:
- each axis has an `id`, which must be unique and by default is `x` for a horizontal axis and `y` for a vertical axis
- use the "position" prop to determine where the axis is
- validate that the series specifies xAxis (default `x`) and yAxis (default `y`) correctly (orthogonal)
- throw if 2 axes have the same position

maasencioh:
If two axes have the same id doesn't throw if are in the same direction:
```tsx
{/* This displays the same axis on the left and right position */}
<Axis position='left' id='y' />
<Axis position='right' id='y' />
```

targos:
I like option 1, because it allows to customize some props differently for each position.

targos:
We need to think about how this will work on the PlotObject.

maasencioh:
In the first case, the key of the PlotObject could be the position, and for the second case the id. Another way is to make the object an array of objects.

targos:
+1 to an array of objects. The Plot component could transform the JSX to that as well so we have the same implementation.

PR #64, opened 2021-01-08
Axis react

maasencioh:
## Changes
- Migrates implementation on Axis using react to modify the DOM
- Merge both Axis and display depending on the position (currently only accepts 'vertical' and 'horizontal')
- Simplify reducers with immerjs

## Reason
This preparates the work for #54, #59, and #60

### DIFF
--- a/src/Plot.tsx
+++ b/src/Plot.tsx
@@ -1,8 +1,7 @@
 import { max, min } from 'd3-array';
 import { scaleLinear, scaleOrdinal } from 'd3-scale';
 import { schemeSet1 } from 'd3-scale-chromatic';
-// eslint-disable-next-line import/no-named-as-default
-import produce from 'immer';
+import { produce } from 'immer';
 import React, { useMemo, useReducer } from 'react';
 
 import { PlotContext, DispatchContext } from './hooks';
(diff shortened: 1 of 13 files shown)


src/Plot.tsx
```diff
@@ -1,6 +1,8 @@
 import { max, min } from 'd3-array';
 import { scaleLinear, scaleOrdinal } from 'd3-scale';
 import { schemeSet1 } from 'd3-scale-chromatic';
+// eslint-disable-next-line import/no-named-as-default
+import produce from 'immer';
```
targos, line 5:
```suggestion
import { produce } from 'immer';
```

src/Plot.tsx
```diff
@@ -22,41 +24,56 @@ interface State {
 }
 
 function reducer(state: State, action: ReducerActions): State {
-  switch (action.type) {
-    case 'newData': {
-      const { series } = state;
-      const serieItem = action.value;
-      return { ...state, series: [...series, serieItem] };
+  return produce(state, (draft) => {
```
targos, line 27: instead of always returning `produce()` you can wrap the reducer function with `produce(function reducer(){})`

stories/plot-object.stories.tsx
```diff
@@ -16,8 +16,13 @@ export function Control() {
       margin: { bottom: 50, left: 50, top: 5, right: 10 },
     },
     axes: {
-      x: { label: 'My X axis', position: 'bottom', min: 0, max: 10 },
-      y: { label: 'My Y axis', position: 'left', min: 0, max: 10 },
+      x: {
+        label: 'My X axis',
+        position: 'horizontal' as const,
```
targos, line 21: what happens without the `as const` ?

maasencioh: something like `string is not assignable to 'horizontal' | 'vertical'`

targos: we need to find another way. I don't want to have to do this everytime I use the object API

maasencioh: what do you think of using enums?

```ts
const enum Positions {
  horizontal = 'horizontal',
  vertical = 'vertical',
}

{ y: { position: Positions.vertical } }
```

src/Plot.tsx
```diff
@@ -91,7 +87,11 @@ export default function Plot({
     xFlip: false,
     yFlip: false,
   };
-  const [state, dispatch] = useReducer(reducer, initialState, undefined);
+  const [state, dispatch] = useReducer(
+    produce(reducer),
```
targos, line 91: Can you keep this call outside of the component to avoid creating a new function on every render?

stories/plot-object.stories.tsx
targos: strings are fine. I think you just need to specify the type when you create the plot object: `const plot: PlotObjectType = { ... }`

src/PlotObject.tsx
```diff
@@ -4,24 +4,15 @@ import Plot from './Plot';
 import Axis from './components/Axis';
 import Legend from './components/Legend';
 import LineSeries from './components/LineSeries';
-import { AxisProps, LegendProps, LineSeriesProps, PlotProps } from './types';
+import { PlotObjectProps } from './types';
```
targos, line 7: not a big fan of this name, because it implies that it's the type of the props for the `PlotObject` component.

targos (approved):

PR #66, opened 2021-01-11
Multiple axis

### LINKED ISSUE #60: Replace XAxis/YAxis with Axis

maasencioh:
## Features
- [x] add scatter plot to PlotObject options
- [x] each axis has an `id`, which must be unique
- [x] use the "position" prop to determine where the axis is
- [x] throw if 2 axes have the same position
- [x] each axis has its own scale and derived state
- [x] validate that the series specifies xAxis (default `x`) and yAxis (default `y`) correctly (orthogonal)
- [x] by default is `x` for a horizontal axis and `y` for a vertical axis

closes #60

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,3 +1,3 @@
 extends: '@zakodium/eslint-config/typescript-react'
 rules:
-  '@typescript-eslint/no-dynamic-delete': 0
+  '@typescript-eslint/no-dynamic-delete': off
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,1 +1,3 @@
 extends: '@zakodium/eslint-config/typescript-react'
+rules:
+  '@typescript-eslint/no-dynamic-delete': 0
--- a/src/Plot.tsx
+++ b/src/Plot.tsx
@@ -5,22 +5,13 @@ import { produce } from 'immer';
 import React, { useMemo, useReducer } from 'react';
 
 import { PlotContext, DispatchContext } from './hooks';
-import type { PlotProps, SeriesType, ReducerActions } from './types';
-import { splitChildren } from './utils';
-
-interface State {
-  series: SeriesType[];
-  xMin?: number;
-  xMax?: number;
-  yMin?: number;
-  yMax?: number;
-  paddingLeft: number;
-  paddingRight: number;
-  paddingTop: number;
-  paddingBottom: number;
-  xFlip: boolean;
-  yFlip: boolean;
-}
+import type {
+  PlotProps,
+  State,
+  ReducerActions,
+  AxisContextType,
+} from './types';
+import { splitChildren, validatePosition } from './utils';
 
 function reducer(state: State, action: ReducerActions) {
   switch (action.type) {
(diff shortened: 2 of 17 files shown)


src/Plot.tsx
```diff
@@ -34,39 +25,26 @@ function reducer(state: State, action: ReducerActions) {
...
-    case 'flip': {
-      const { flip, axis } = action.value;
-      if (axis === 'x') {
-        state.xFlip = flip;
-      } else {
-        state.yFlip = flip;
-      }
+    case 'removeAxis': {
+      const { id } = action.value;
+      const seriesFiltered = state.series.filter((series) => series.id !== id);
```
targos, line 41: that seems weird to filter the series on an action called "removeAxis" ?

maasencioh: Oups, good catch

src/Plot.tsx
```diff
@@ -38,8 +38,7 @@ function reducer(state: State, action: ReducerActions) {
     }
     case 'removeAxis': {
       const { id } = action.value;
-      const seriesFiltered = state.series.filter((series) => series.id !== id);
-      state.series = seriesFiltered;
+      state.axis[id] = undefined;
```
targos, line 41:
```suggestion
      delete state.axis[id];
```

maasencioh: wasn't it more performant the undefined assignation?

targos: Performance doesn't matter much here, as we are not really modifying an object but actually calling `immer`'s Proxy (in both cases). I find delete more explicit.

maasencioh: that will [break a rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-dynamic-delete.md), what do you think of using a Map instead?

targos: https://www.youtube.com/watch?v=Xi5rxgDA0Pg
We can disable this rule in the project. Using a Map sounds good too. Just check how to enable it (not globally) in immer as I think it's not enabled by default.

targos: Somehow my answer was not sent yesterday, sorry.
I'm OK with using a Map, but be careful because immer doesn't support it by default, you need to enable it: https://immerjs.github.io/immer/docs/installation#pick-your-immer-version
I'm also OK with disabling that ESLint rule in the project: https://www.youtube.com/watch?v=Xi5rxgDA0Pg

maasencioh: I'll go for disabling the rule in this case https://youtu.be/L397TWLwrUU

.eslintrc.yml
```diff
@@ -1 +1,3 @@
 extends: '@zakodium/eslint-config/typescript-react'
+rules:
+  '@typescript-eslint/no-dynamic-delete': 0
```
targos, line 3:
```suggestion
  '@typescript-eslint/no-dynamic-delete': off
```

targos (approved):

PR #68, opened 2021-01-13
feat: allows to hide series and specify serie id

### LINKED ISSUE #49: Marker API in LineSeries vs. ScatterSeries
Should we have two way to draw markers?
Should both ways support the same options (i.e. should LineSeries expose **all** options from ScatterSeries)?

maasencioh:
- hide scatter or line series
- optional id (adds the same color to line and scatter series)
- line series extends scatter series

closes #49

### DIFF
--- a/stories/examples/iris-dataset.stories.tsx
+++ b/stories/examples/iris-dataset.stories.tsx
@@ -70,6 +70,7 @@ export function PCA() {
               xAxis="x"
               yAxis="y"
               key={klass}
+              id={klass}
               data={data}
               label={klass}
               markerSize={2}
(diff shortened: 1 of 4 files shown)


maasencioh:
Here the [PCA example](https://deploy-preview-68--amazing-kilby-d9c1d7.netlify.app/?path=/story/examples-iris-dataset--pca) will have the same color for the data and the regression line

stories/examples/iris-dataset.stories.tsx
```diff
@@ -80,6 +81,7 @@ export function PCA() {
               xAxis="x"
               yAxis="y"
               key={klass}
+              id={klass}
```
targos, line 84: can we find a better name than "id" ? For me, "id" implies that it cannot be the same for different elements

targos (approved):

PR #70, opened 2021-01-13
feat: add option for getting ticks inside the axis

### LINKED ISSUE #54: support putting ticks inside the plot

maasencioh:
closes #54

### DIFF
--- a/src/components/Axis/BottomAxis.tsx
+++ b/src/components/Axis/BottomAxis.tsx
@@ -14,6 +14,7 @@ export default function BottomAxis({
   fontSize,
   tickStyle,
   showTicks,
+  tickInside,
 }: AxisChildProps) {
   const {
     axisContext,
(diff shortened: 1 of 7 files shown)


stropitek:
What about a prop that sets the tick length, so that you can both control the size and the position inside / outside with one prop?

maasencioh:
@stropitek I thought the same, but then if you want to invert it you should add a negative value, that's not very user friendly right?

stropitek:
Yes fair enough, it would have to be properly documented. But maybe better to have size just control the size.

stropitek:
I'm not sure if the tick direction option should be a boolean or a string.

`<Axis tickInside />` vs `<Axis tickDirection="inside" />` with the default value being `outside`

In that case a boolean would make sense because there are only 2 possibilities but I still find the first one a bit awkward compared to the second one. But I could go either way.

stropitek (approved):

maasencioh:
Maybe `tickEmbedded`, following the labels name convention?

stropitek:
> Maybe `tickEmbedded`, following the labels name convention?

yes that is better than `tickInside` IMO

Issue #71, opened 2021-01-15
Add IR storybook


lpatiny:
In the storybook there should be a new section called 'Experimental spectra'

First example is here and is called 'Infrared'

![image](https://user-images.githubusercontent.com/1484241/104708063-814e6a00-571d-11eb-930e-0543dc87ff4d.png)

[ir.json.zip](https://github.com/zakodium/react-plot/files/5819615/ir.json.zip)

Please take care about inverted horizontal axis and put the unit between square brackets (cm-1). Also the '-1' has to be superscript but this can be found in UTF8 I think without you have to add a styling option.

PR #81, opened 2021-01-15
Feat spectra

Sebastien-Ahkrin:
close: #71


lpatiny:
For all that is scientific data we will try to be rigorous with units and systematically put it between square brackets

So for for X axis: Wavelength [cm-1]
For Y axis: Transmittance [%]

We will also try to start always with an uppercase the axis labels

targos:
We should not use `feat: ` for storybook changes because it doesn't change the public API.
`docs: ` or `test: ` can be used for that.

PR #83, opened 2021-01-18
Feat add examples

Sebastien-Ahkrin:
NMR:
image
 
RAMAN:
image

For the Raman example, we should hide the number in the yAxis but actually we do not have any props to do this (see #82)


maasencioh:
@Sebastien-Ahkrin a workaround in this to hide the axis and wrap the plot with the viewport style

maasencioh (approved):

PR #86, opened 2021-01-18
refactor: change the implementation of Tick in Axis

Sebastien-Ahkrin:


src/components/Ticks.tsx
```diff
@@ -0,0 +1,85 @@
+import { ScaleLinear } from 'd3-scale';
+import React, { CSSProperties, ReactNode, SVGAttributes } from 'react';
+
+interface NumberXY {
```
targos, line 4:
```suggestion
interface CoordinatesXY {
```

src/components/Ticks.tsx
```diff
@@ -0,0 +1,85 @@
...
+  anchor?: SVGAttributes<SVGTextElement>['textAnchor'];
+  alignment?: SVGAttributes<SVGTextElement>['alignmentBaseline'];
+}
+
+export interface TicksProps extends Omit<TickProps, 'line' | 'text'> {
+  show: boolean;
+  ticks: number[];
+  scale: ScaleLinear<number, number, never>;
+
+  getValues: (
```
maasencioh, line 28:
```suggestion
  getPositions: (
```

maasencioh (approved):
Naming is always something that creates division

Issue #75, opened 2021-01-15
Superimpose 2 TGA spectra


lpatiny:
In story book spectra add in the section 'spectrum' the superimposition of those 2 spectra

Y axis should be forced to start at 0 and end at 1

![image](https://user-images.githubusercontent.com/1484241/104712864-67b02100-5723-11eb-8d64-68f3f9c73b16.png)

[Archive.zip](https://github.com/zakodium/react-plot/files/5819840/Archive.zip)

PR #90, opened 2021-01-19
Refactor feat #75

Sebastien-Ahkrin:
close: #75


src/PlotReducer.tsx
```diff
@@ -0,0 +1,37 @@
+import { ReducerActions, State } from './types';
```
maasencioh, line 1: by convention this file should be called PlotReducer.ts, tsx is for jsx with typescript

src/types.ts
```diff
@@ -135,9 +135,11 @@ export interface PlotContextType {
   axisContext: Record<string, AxisContextType>;
 }
 
+export type AxisType = Record<string, Omit<AxisParentProps, 'id'>>;
```
maasencioh, line 138: is this type used somewhere else? if not you could skip the exportation

stories/spectra/tga.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+      width={1400}
+      height={540}
+      margin={{ bottom: 45, left: 40, top: 40, right: 40 }}
+      viewportStyle={{ stroke: 'black' }}
+    >
+      <LineSeries
+        data={data1}
+        xAxis="x"
+        yAxis="y"
+        lineStyle={{ stroke: 'red' }}
```
maasencioh, line 24: you could remove both strokes, by default is giving these colors https://github.com/d3/d3-scale-chromatic#schemeSet1

targos (approved):

maasencioh (commented):
At the end of the blue curve you have a curve, is this supposed to be correct?
Screenshot 2021-01-20 100651

src/PlotReducer.tsx
targos: > At the end of the blue curve you have a curve, is this supposed to be correct?
> Screenshot 2021-01-20 100651

Yes, it is how the data is.

targos:
> At the end of the blue curve you have a curve, is this supposed to be correct?
> Screenshot 2021-01-20 100651

Yes, it is how the data is.

PR #93, opened 2021-01-20
Sebastien ahkrin/issue88

Sebastien-Ahkrin:


src/types.ts
```diff
@@ -48,13 +48,13 @@ export interface AxisChildProps {
...
-  showGridLines?: boolean;
+  displayGridLines?: boolean;
   labelStyle?: CSSProperties;
   tickStyle?: CSSProperties;
   tickEmbedded?: boolean;
   labelSpace?: number;
-  display?: boolean;
-  showTicks?: boolean;
+  hidden?: boolean;
+  hiddenTicks?: boolean; // showTicks = true => hiddenTicks = false
```
targos, line 57:
```suggestion
  hiddenTicks?: boolean;
```

targos (approved):

targos:
Don't worry about the additional commits. They will be squashed and are already on master.

src/components/Ticks.tsx
```diff
@@ -78,7 +77,7 @@ export function Tick(props: TickProps) {
         x={textX1}
         y={textY1}
         textAnchor={anchor}
-        alignmentBaseline={alignment}
+        dominantBaseline="middle"
```
maasencioh, line 80: If you look at the bottom axis, the texts are lower, is it for this change?

targos: As said in my previous comment, this is already on master

maasencioh (approved):

stropitek:
Shouldn't we choose imperative style for all boolean props? `show`, `hide`, `showTicks`, `hideTicks`

What I've seen in other libraries (I looked in nivo) is that when using an adjective or past participle, the `is` prefix is added. They also have nouns. Example in nivo: `animate`, `enableLabel`, `isInteractive`, `leavesOnly`

stropitek:
After discussing with Michael, we came to the conclusion that it is hard to find a general rule that also makes the all the options feel natural.

I also find that `hidden` is a good prop name.

PR #99, opened 2021-01-20
Zoom

maasencioh:
This is the first approach to make a zoom component over the plot.
Scroll over the plot updates the selected positions (left and bottom are default) padding.

https://deploy-preview-99--amazing-kilby-d9c1d7.netlify.app/?path=/story/plot-plot-zoom--control


maasencioh:
This is going to be implemented in another way and another moment

PR #100, opened 2021-01-21
docs: add new example 'Nasqad Simulation' in Storybook

### LINKED ISSUE #89: Simulate nasdaq value
Have a storyboard that would update the chart every second based on a simulated nasdaq value.

On the horizontal axis we would have the time, on the vertical axis 'Simulated Nasdaq value [USD]'

The plot would always contain the last 500s and would be refreshed every second.

Would be nice if the data 'looks' a little bit real with some reasonable oscillation.

NadjibSb:
Closes #89


NadjibSb:
For date & time axis , Should we do something like we did for scientific values ?

If I may, I suggest to give the user the possibility to handle the display of axis values by passing a callback (it's better than editing the data itself)

stories/spectra/nasqad.stories.tsx
```diff
@@ -0,0 +1,91 @@
+import { Meta } from '@storybook/react';
+import React, { useState, useEffect } from 'react';
+
+import { Axis, LineSeries, Plot, Heading } from '../../src';
+import type { Series } from '../../src/types';
+import srcData from '../data/nasdaq.json';
+
+export default {
+  title: 'Experimental spectra/Nasqad',
```
lpatiny, line 9: Please check everywhere so that it is Nasdaq and not Nasqad

stories/spectra/nasqad.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const [data, setData] = useState(srcData);
+
+    useEffect(() => {
+      const timer = setInterval(() => {
+        const [x, y] = generateNewXY(data[data.length - 1], step); // generate a new conrdinates
+        const newData = getLastData([...data, { x, y }], displayInterval); // get last 500s data
+        setData(newData);
+      }, refreshFrequency * 1000);
+      return () => clearInterval(timer);
+    });
```
lpatiny, line 58: Do you have eslint plugin in vscode ? Normally there should be a warning to enforce the specification of the variables to listen.
So please add `[data]`

stories/spectra/nasqad.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+    refreshFrequency, // Page refresher frequency (s)
+    step, // timestamp's step
+    displayInterval, // interval to display (s)
+  } = props;
+
+  function PlotRefresher() {
+    const [data, setData] = useState(srcData);
+
+    useEffect(() => {
+      const timer = setInterval(() => {
```
lpatiny, line 52: In this case you could use setTimeout (and cancel the timeout in the return). Anyway it will never be called more than once (if I understand well).

lpatiny (commented):
Are those real data ?

stories/spectra/nasqad.stories.tsx
targos: Also please move this to the "Examples" category. This data is not an experimental spectrum

stories/spectra/nasqad.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+        const [x, y] = generateNewXY(data[data.length - 1], step); // generate a new conrdinates
```
targos, line 53:
```suggestion
        const [x, y] = generateNewXY(data[data.length - 1], step); // generate a new coordinates
```

stories/spectra/nasqad.stories.tsx
NadjibSb: Exactly,  
setTimeout will never be called more than once, which is not what we aim for.
setInterval allows us to run a function repeatedly, starting after the interval of time, then repeating continuously at that interval.

targos: Yes, setInterval allows us to run the function repeatedly, but this is not what happens here, because everytime the `setData` function is called, the component is re-rendered, the `useEffect` cleanup function is executed, clearing the interval and a new interval is created. So in practice, setTimeout has the same effect in this situation.

NadjibSb: You are right, 
setTimeout is a better fit since the hook cleanup the timer after every render

targos:
Please rebase on the master branch. We changed the API in https://github.com/zakodium/react-plot/commit/ae2991a4f9c443d01d6f9fdab2c4dcdd3394484c

NadjibSb:
> Are those real data ?

No it's not, I generated it myself with a reasonable oscillation

PR #102, opened 2021-01-21
refactor: rename show in Ticks props

### LINKED ISSUE #101: Change show prop in Tick
use `hidden` instead of `show` like elsewhere

Sebastien-Ahkrin:
Fixes #101

### DIFF
--- a/src/components/Axis/BottomAxis.tsx
+++ b/src/components/Axis/BottomAxis.tsx
@@ -67,7 +67,7 @@ export default function BottomAxis({
           <Ticks
             scientific={scientific}
             scale={scale}
-            show={!hiddenTicks}
+            hidden={!hiddenTicks}
             ticks={ticks}
             style={tickStyle}
             anchor="middle"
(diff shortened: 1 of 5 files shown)


targos:
Hidden is the opposite of show. You need to change the logic too

src/components/Axis/Ticks.tsx
```diff
@@ -33,14 +33,14 @@ export interface TicksProps extends Omit<TickProps, 'line' | 'text'> {
 
 export function Ticks(props: Omit<TicksProps, 'children'>) {
   const {
-    show,
+    hidden = false,
```
targos, line 36:
```suggestion
    hidden,
```

The prop is mandatory

src/components/Axis/BottomAxis.tsx
```diff
@@ -67,7 +67,7 @@ export default function BottomAxis({
           <Ticks
             scientific={scientific}
             scale={scale}
-            show={!hiddenTicks}
+            hidden={!hiddenTicks}
```
targos, line 70: This still seems wrong (same in the other ...Axis components)

maasencioh (approved):

PR #128, opened 2021-01-27
feat: add secondary ticks 

### LINKED ISSUE #116: Secondary ticks
Give the possibility to allow to draw secondary ticks. Those ticks will not be labelled but we could style them.

NadjibSb:
closes #116

### DIFF
--- a/src/components/Axis/Ticks.tsx
+++ b/src/components/Axis/Ticks.tsx
@@ -27,7 +27,7 @@ export interface TicksProps extends Omit<TickProps, 'line' | 'text'> {
   ticks: number[];
   scale: ScaleLinear<number, number, never>;
   scientific?: boolean;
-  secondaryTicksHidden?: boolean;
+  hiddenSecondaryTicks?: boolean;
 
   getPositions: (
     y: number,
@@ -41,7 +41,7 @@ export function Ticks(props: Omit<TicksProps, 'children'>) {
     scale,
     getPositions,
     scientific = true,
-    secondaryTicksHidden = false,
+    hiddenSecondaryTicks = false,
     ...otherProps
   } = props;
   if (hidden) return null;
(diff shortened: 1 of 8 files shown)


NadjibSb:
#116 
Here is some results : 

image
image
image

The secondary Ticks expend according to the density of the main ticks, & they are intuitive to the reader

targos:
I force-pushed to remove the merges

src/components/Axis/Ticks.tsx
```diff
@@ -10,12 +10,14 @@ interface CoordinatesXY {
...
   style: CSSProperties;
-  children: ReactNode;
+  children?: ReactNode;
 
   line: CoordinatesXY;
   text: Omit<CoordinatesXY, 'x2' | 'y2'>;
+  secondary?: boolean;
 
   strokeColor?: string;
+  strokeHight?: number;
```
lpatiny, line 20: Small spelling mistake. Should be strokeHeight. I'm using a plugin in vscode : 'Code Spell Checker' to highlight those kind of errors.

src/utils.ts
```diff
@@ -128,16 +128,14 @@ export function functionalStyle<T>(
 export function calculateTicksNumber(
   plotWidth: number,
   scientific: boolean,
-  domaine: number[] | undefined,
+  domaine: number[] = [0, 1],
```
lpatiny, line 131: Seems to me in english it should be 'domain'

src/components/Axis/Ticks.tsx
```diff
@@ -25,6 +27,7 @@ export interface TicksProps extends Omit<TickProps, 'line' | 'text'> {
   ticks: number[];
   scale: ScaleLinear<number, number, never>;
   scientific?: boolean;
+  secondaryTicksHidden?: boolean;
```
maasencioh, line 30: Because the name of the main feature is `hiddenTicks`, what do you think of naming the variable like this?
```suggestion
  hiddenSecondaryTicks?: boolean;
```

PR #129, opened 2021-01-28
chore: throw error if width or height are undefined

maasencioh:
otherwise, it will silently fail and add `NaN` values to the SVG, and plotting only the axis

image

### DIFF
--- a/src/Plot.tsx
+++ b/src/Plot.tsx
@@ -33,6 +33,9 @@ export default function Plot({
   if (hasInvalidChild) {
     throw new Error('Only compound components of Plot are displayed');
   }
+  if ([width, height].includes(undefined)) {
+    throw new Error('Width and heigth are mandatory');
+  }
 
   // Distances in plot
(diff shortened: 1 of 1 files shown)


stropitek:
Can this also happen with typescript?

maasencioh:
@stropitek `common-spectrum` is not using typescript and is the one that generates the object for the `PlotObject` component. I'll migrate `common-spectrum` to typescript, but after working on some other priorities with Luc

src/Plot.tsx
```diff
@@ -33,6 +33,9 @@ export default function Plot({
   if (hasInvalidChild) {
     throw new Error('Only compound components of Plot are displayed');
   }
+  if ([width, height].includes(undefined)) {
+    throw new Error('Width and heigth are mandatory');
```
stropitek, line 37:
```suggestion
    throw new Error('Width and height are mandatory');
```

stropitek (approved):

PR #130, opened 2021-01-28
feat: add bar series component

### LINKED ISSUE #77: Add a BarSeries component
Needed to implement https://github.com/zakodium/react-plot/issues/74

It will draw a vertical line from 0 for each point.

Sebastien-Ahkrin:
Fixes #77

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,13 +1,12 @@
 import React, { CSSProperties } from 'react';
 
 import Axis from './components/Axis';
+import BarSeries from './components/BarSeries';
 import Heading from './components/Heading';
 import Legend from './components/Legend';
 import LineSeries from './components/LineSeries';
 import ScatterSeries from './components/ScatterSeries';
 import type { AxisContextType, CSSFuncProps, PlotChildren } from './types';
-
-import { BarSeries } from '.';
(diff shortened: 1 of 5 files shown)


stories/control/line.stories.tsx
```diff
@@ -1,7 +1,15 @@
...
+  Legend,
+  LineSeries,
+  LineSeriesProps,
+  Plot,
+  BarSeries,
+} from '../../src';
+import { DEFAULT_CONFIG } from '../utils';
 
 export default {
   title: 'API/LineSeries',
```
stropitek, line 15: The bar series probably deserves it's own title even if it shares the same API for now

targos: It should have its own story file too

src/utils.ts
```diff
@@ -7,6 +7,8 @@ import LineSeries from './components/LineSeries';
 import ScatterSeries from './components/ScatterSeries';
 import type { AxisContextType, CSSFuncProps, PlotChildren } from './types';
 
+import { BarSeries } from '.';
```
targos, line 10: please import it from the file directly to avoid a circular dependency

PR #137, opened 2021-02-02
feat: add Annotation

Sebastien-Ahkrin:


maasencioh:
What do you think if for the arrow you make the triangle at the end aligned to the direction on the line? having the starting and ending point you can calculate the angle and make the rotation (it should be something like `rotate(${Math.atan(dx/dy)}rad)`

src/Plot.tsx
```diff
@@ -95,6 +102,8 @@ export default function Plot(props: PlotProps) {
             height={height}
             style={{ ...defaultSvgStyle, ...svgStyle }}
           >
+            <MarkerDefs />
```
maasencioh, line 105: If is only used for the arrows, shouldn't be inside de arrow annotation component? or it needs to be in there?

stropitek: It's to avoid repeating the same definitions many times again. It's small and we can optimize and not add it if no annotations are rendered.

stropitek:
> What do you think if for the arrow you make the triangle at the end aligned to the direction on the line?

It's already the case thanks to the `marker` attribute `orient`

stories/annotations.stories.tsx
```diff
@@ -0,0 +1,192 @@
...
+    x: 3,
+    y: 16,
+  },
+  {
+    x: 4,
+    y: 18,
+  },
+];
+
+export function AnnotationRectangleStories() {
```
stropitek, line 72: Can you remove the stories suffix from all exported stories?

They end up in the story name

Sebastien-Ahkrin: i have updated the stories, can you say to me if its good for you ?

stropitek (approved):

PR #141, opened 2021-02-03
feat: add error bars 

### LINKED ISSUE #133: Possibility to add error bars
This applied to all Series (Scatter, Line)

Extend the data type to allow new optional number or array properties
- xError: this value will be used to draw vertical error bars
- yError: this value will be used to draw horizontal error bars

We may have the following example:
- xError:1 : symmetric error bar
- xError:[null, 0.7] : only one error bar on toward the high value
- yError: [1,2]: one error bar of 1 to the low value, one error bar of 2 to the high value

Take care to make the distinction between 0 and undefined. An error may be 0 and the error bar should still be drawn (actually only the cap). If it is undefined (or null) the error bar should be drawn. In order to check if the variable is undefined or null you should use `x == null`. The only exception where we are allowed to use `==` and not `===` ...

Error bars are drawn if the property `errorBars` is defined that has various properties.
```
errorBars: {
  style: {},
  capSize: 4, // in pixels
  capStyle: {}
}
```

Example of chart:

image

NadjibSb:
fixes #133


src/components/ErrorBars.tsx
```diff
@@ -0,0 +1,150 @@
...
+    right,
+    style,
+    capSize = 10,
+    capStyle,
+  } = props;
+  const defaultColor = 'black';
+
+  return (
+    <g>
+      {!(top == null) && (
```
maasencioh, line 65:
```suggestion
      {top && (
```

src/components/ErrorBars.tsx
```diff
@@ -0,0 +1,150 @@
...
+            x2={origin.x + capSize / 2}
+            y1={top}
+            y2={top}
+            stroke={defaultColor}
+            {...capStyle}
+          />
+        </g>
+      )}
+
+      {!(bottom == null) && (
```
maasencioh, line 86:
```suggestion
      {bottom && (
```

src/components/ErrorBars.tsx
```diff
@@ -0,0 +1,150 @@
...
+            x2={origin.x + capSize / 2}
+            y1={bottom}
+            y2={bottom}
+            stroke={defaultColor}
+            {...capStyle}
+          />
+        </g>
+      )}
+
+      {!(left == null) && (
```
maasencioh, line 107:
```suggestion
      {left && (
```

src/components/ErrorBars.tsx
```diff
@@ -0,0 +1,150 @@
...
+            x2={left}
+            y1={origin.y - capSize / 2}
+            y2={origin.y + capSize / 2}
+            stroke={defaultColor}
+            {...capStyle}
+          />
+        </g>
+      )}
+
+      {!(right == null) && (
```
maasencioh, line 128:
```suggestion
      {right && (
```

stories/control/line.stories.tsx
```diff
@@ -36,6 +36,41 @@ export default {
...
     },
+    // ErrorBars props
+    hiddenErrorBars: {
+      control: 'boolean',
+      defaultValue: false,
+      table: {
+        category: 'Error Bars',
+      },
+    },
+    ErrorBarsStyle: {
```
lpatiny, line 47: First letter should be lowerCase

lpatiny:
Please could you check why when the full Series is hidden the error bars are still displayed ? There may be some design error somewhere.

image

Also could you add one more example dedicated to error bars in the storybook ?

NadjibSb:
> Please could you check why when the full Series is hidden the error bars are still displayed ? There may be some design error somewhere.
> 
> image
> 
> Also could you add one more example dedicated to error bars in the storybook ?

@lpatiny should I hide the errorBars when the marker is hidden as well (for Line series) ?

stropitek (commented):
The `errorBarStyle` and `capStyle` props in the story cannot be edited. Could you please fix this?

Can you also merge `style` and `capStyle` to style the caps : `<line {...style} {...capStyle} />` so that capStyle needs only to be used to override the main style. This make the api easier in th 99% of cases where we don't need a different style on the cap.

src/components/ErrorBars.tsx
NadjibSb: @maasencioh  I didn't do that cuz when top==0 the line should still be drawn

NadjibSb:
> The `errorBarStyle` and `capStyle` props in the story cannot be edited. Could you please fix this?

It's due to that Storybook don't support controls for complex objects.
I looked it up, & apparently the issue is still open : 
check : [storybookjs/storybook/#12078 ](https://github.com/storybookjs/storybook/issues/12078) & [storybookjs/storybook/#12362](https://github.com/storybookjs/storybook/issues/12362)

src/components/ErrorBars.tsx
targos: The please use the `!=` operator

src/components/LineSeries.tsx
```diff
@@ -65,10 +65,11 @@ export default function LineSeries(props: LineSeriesProps) {
     <g>
       {props.hidden ? null : <LineSeriesRender {...lineProps} />}
       <ErrorBars
+        {...props.errorBars}
         xAxis={props.xAxis}
         yAxis={props.yAxis}
         data={props.data}
-        {...props.errorBars}
+        hidden={props.errorBars?.hidden || props.hidden}
```
targos, line 72: can you please keep the original spread and instead put the "props.hidden" condition one level higher ? i.e. render `null` instead of the group. Then you can simplify <LineSeriesRender and <ScatterSeries too.

NadjibSb:
Fot the data format, We are working like @lpatiny suggested :
- x: 5, xError: [ 0.2 , 0.1 ]

But I think the Error should be specified as : 
- x: 5, xError: [ -0.2 , 0.1 ]
 Or
- x: 5, xError: [ 4.8 , 5.1 ]

stropitek:
The name of the props in the story should reflect the API otherwise it is misleading, especially since the chapter is calle "API"

I discussed with Michael and we suggest to flatten the props for the error bar, as it is already the case for the line serie markers, and use `displayErrorBars` instead of `hidden` in the nested prop.
```jsx
<LineSerie
  displayErrorBars
  errorBarsStyle
  errorBarsCapStyle
  errorBarsCapSize
/>
```

This will also resolve the issue with the destructuring of props with storybook  object controls

lpatiny:
> @lpatiny should I hide the errorBars when the marker is hidden as well (for Line series) ?

No. We may want to display error bars without marker.

lpatiny:
> * x: 5, xError: [ -0.2 , 0.1 ]
>   Or
> * x: 5, xError: [ 4.8 , 5.1 ]

Errors are usually calculate with a square root of the sum of squares of difference... Just to say that it is a positive number. So we would prefer to keep  [ 0.2 , 0.1 ].

What was the reason you wanted to change ?

NadjibSb:
> What was the reason you wanted to change ?

When I was looking in some datasets to add an example, I found some data following the format I suggested. But I think it really depends on the context & data source.

src/components/LineSeries.tsx
```diff
@@ -60,13 +60,34 @@ export default function LineSeries(props: LineSeriesProps) {
...
     <g>
-      {props.hidden ? null : <LineSeriesRender {...lineProps} />}
+      <LineSeriesRender {...lineProps} />
+      <ErrorBars {...errorBarsProps} />
       <ScatterSeries
         {...otherProps}
-        hidden={!displayMarker || props.hidden}
+        hidden={!displayMarker}
         groupId={id}
+        displayErrorBars={false}
```
stropitek, line 90: we usually do not specify boolean props unless their value needs to be `true`
```suggestion
        displayErrorBars={false}
```

stories/examples/bitcoin.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+        subtitle={'With High & Low prices'}
+      />
+      <LineSeries
+        data={data}
+        xAxis="x"
+        yAxis="y"
+        lineStyle={{ stroke: 'green', strokeWidth: 2 }}
+        displayMarker={true}
+        markerStyle={{ fill: 'green', stroke: 'none' }}
+        displayErrorBars={true}
```
stropitek, line 33: we usually pass boolean props with the shorter syntax

```suggestion
        displayErrorBars
```

stropitek (approved):

stropitek:
@NadjibSb have all comment been adressed or are you still working on something in the PR?

NadjibSb:
> @NadjibSb have all comment been adressed or are you still working on something in the PR?

Yes, all comment have been adressed.

PR #142, opened 2021-02-03
fix: Tick text aligned according to ticks length

### LINKED ISSUE #139: Tick text is not aligned according to the tickLenght
When tickLength > 6 , the text overlap with the tick.
Ex: 

image

NadjibSb:
fixes #139


maasencioh (approved):
I think that this is going to be fixed anyway in `react-d3-utils`

PR #145, opened 2021-02-26
fix: remove legend on unmount

maasencioh:
- [x] unmount label when a series is removed
- [x] same state when series is added again

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "singleQuote": true
   },
   "devDependencies": {
+    "@simbathesailor/use-what-changed": "^1.0.0",
     "@storybook/addon-essentials": "^6.1.17",
     "@storybook/addon-links": "^6.1.17",
     "@storybook/addon-storysource": "^6.1.17",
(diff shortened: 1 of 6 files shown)


targos:
should we wait for you to do the second item in the list before reviewing?

targos:
Please add a story to show the fix.

maasencioh:
It's already in this story, you could use the control `displayPlot`

https://deploy-preview-145--amazing-kilby-d9c1d7.netlify.app/?path=/story/plot-general-options--control

package.json
```diff
@@ -45,6 +45,7 @@
     "singleQuote": true
   },
   "devDependencies": {
+    "@simbathesailor/use-what-changed": "^1.0.0",
```
targos, line 48: this seems unused

maasencioh: I used it for debugging the `useEffect`, it could be useful for future debugging processes

targos: can you add something about it in the readme then? maybe with a link to the documentation?
otherwise nobody except you know why it's in the dependencies

maasencioh: thanks

PR #149, opened 2021-03-11
feat: add Range Series component

Sebastien-Ahkrin:


stories/control/range.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+  component: RangeSeries,
+  argTypes: {
+    displayMarker: {
+      control: 'boolean',
+      defaultValue: true,
+    },
+    lineStyle: {
+      control: 'object',
+      defaultValue: {
+        stroke: 'black',
```
targos, line 16:
```suggestion
        stroke: 'black',
        fill: 'grey',
```

targos:
I see you added ScatterRangeSeries.tsx and ErrorRangeBars.tsx. How do they work? Can you add a story to show the features?

targos:
Please change the LineSeries example in the story so it does not look like it's part of the range, the result is disturbing :D

PR #160, opened 2021-03-26
feat: add logScale

maasencioh:
It needs to be migrated the linear scale to `react-d3-utils` and added the secondary ticks to that library.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -144,7 +144,7 @@ export interface AxisParentProps {
   paddingEnd?: number;
 
   flip?: boolean;
-  logScale?: boolean;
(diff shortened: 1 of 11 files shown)


src/types.ts
```diff
@@ -133,6 +144,7 @@ export interface AxisParentProps {
   paddingEnd?: number;
 
   flip?: boolean;
+  logScale?: boolean;
```
targos, line 147: To prepare for the third scale (time), we had in mind to use a `scale="log"` prop. See https://github.com/zakodium/react-plot/issues/78#issue-786806949

targos (approved):

PR #165, opened 2021-03-29
fix: dont' break on values out of logscale domain

maasencioh:
Before it was not showing any tick, so now it clamps to the minimum positive value on that axis

### DIFF
--- a/src/Plot.tsx
+++ b/src/Plot.tsx
@@ -62,10 +62,7 @@ export default function Plot(props: PlotProps) {
   const plotHeight = height - top - bottom;
 
   // Set scales
-  const axisContext = useAxisContext(state, {
-    plotWidth,
-    plotHeight,
-  });
+  const axisContext = useAxisContext(state, { plotWidth, plotHeight });
 
   const labels = useMemo(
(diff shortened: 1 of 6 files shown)


targos:
What if the data are all negative?

maasencioh:
@targos will be very odd that something like that you will like to plot on the logscale right? this change was more for some edge cases, not for a complete dataset like that.

I guess that at that moment we could try to use a pseudo-log function, but I don't think that should be in this PR

targos:
I'd argue that if your data span both in the negative and positive values, either the data are wrong or they should not be plotted with a log scale.
I don't really like the idea of automatically "fixing" that in react-plot.

maasencioh:
Indeed the values are not meant to be plotted for logscale. In this case, they are considered outliers, but I agree that it shouldn't be in the react-plot where they should be processed. Now, should we throw for both cases or just handle the outliers? The third option is to plot them in a pseudo-log, [like in R](https://rdrr.io/github/paodan/funcTools/src/R/pseudoLog.R).

maasencioh:
This process is delegated to the user

PR #169, opened 2021-04-07
feat: add className to svg

maasencioh:


stropitek:
Please hint to the breaking change in the semantic commit

src/Plot.tsx
```diff
@@ -101,6 +103,8 @@ export default function Plot(props: PlotProps) {
             width={width}
             height={height}
             style={{ ...defaultSvgStyle, ...svgStyle }}
+            id={id}
+            className={className}
```
targos, line 107: should be `svgId` and `svgClassName` for consistency with the already existing `svgStyle`.

src/types.ts
```diff
@@ -167,15 +175,18 @@ export interface MarkersProps {
...
     | ({ type: 'line' } & LineSeriesProps)
     | ({ type: 'scatter' } & ScatterSeriesProps)
   >;
   legend?: LegendProps;
-  dimensions: Dimensions;
+  dimensions?: Pick<PlotProps, 'width' | 'height' | 'margin'>;
+  svg?: Pick<
+    PlotProps,
+    'plotViewportStyle' | 'svgStyle' | 'seriesViewportStyle'
+  > & { class?: string; id?: string };
```
targos, line 189: Here, I'd also rename `svgStyle` to `style` and use `className` because that's more "react"

maasencioh: That was part of the discussion that we have with @stropitek, to have the object definition "framework free" (svgStyle I think that has sense to rename anyway)

targos: even outside of a framework, the DOM property is called `className`

stropitek: I also prefer `svg.style` over `svg.svgStyle`
I'm good with `className` didn't think about how the property was named outside of react

stropitek (approved):

targos (approved):

PR #171, opened 2021-04-13
Tracking

### LINKED ISSUE #167: Tracking
We would consider 2 tracking options for scatter series, linear series, and bar series. The first one will be the closest existing point from all the series (this could be archived using a [Voronoi diagram](https://observablehq.com/@mbostock/the-delaunays-dual)).

The second option is to find the closest X value, considering X as the independent axis. This should return the closest value of all the series at the X-axis. The range series only should implement this case, returning the corresponding limits.

Will be needed to see the values of series when I move the mouse over a position in a plot, giving a [tooltip](https://airbnb.io/visx/docs/tooltip) or visual hint of the tracked position and the requested data, similar to this case https://airbnb.io/visx/xychart

example

maasencioh:
closes #167

### DIFF
--- a/src/Plot.tsx
+++ b/src/Plot.tsx
@@ -4,6 +4,7 @@ import { produce } from 'immer';
 import { CSSProperties, Reducer, useMemo, useReducer } from 'react';
 
 import MarkerDefs from './components/Annotations/MarkerDefs';
+import Tracking, { closestCalculation } from './components/Tracking';
 import TransparentRect from './components/TransparentRect';
 import { LegendProvider } from './components/legendsContext';
 import { PlotContext, DispatchContext, useAxisContext } from './hooks';
@@ -24,7 +25,7 @@ const defaultSvgStyle: CSSProperties = {
 export type { PlotProps };
 
 /**
- * Static plot with fixed dimensions.
+ * Static plot with fixed dimension.
  */
 export default function Plot(props: PlotProps) {
   const {
@@ -37,6 +38,9 @@ export default function Plot(props: PlotProps) {
     svgClassName,
     plotViewportStyle = {},
     seriesViewportStyle = {},
+    onClick,
+    onMouseMove,
+    closest,
     children,
   } = props;
 
@@ -49,7 +53,6 @@ export default function Plot(props: PlotProps) {
     heading,
     legend,
     annotations,
-    tracking,
   } = splitChildren(children);
 
   if (hasInvalidChild) {
@@ -135,7 +138,29 @@ export default function Plot(props: PlotProps) {
               </g>
 
               {axes}
-              {tracking}
+
+              <Tracking
+                onClick={(position) => {
+                  onClick?.(
+                    closestCalculation(
+                      position,
+                      closest,
+                      state.series,
+                      axisContext,
+                    ),
+                  );
+                }}
+                onMouseMove={(position) => {
+                  onMouseMove?.(
+                    closestCalculation(
+                      position,
+                      closest,
+                      state.series,
+                      axisContext,
+                    ),
+                  );
+                }}
+              />
             </g>
 
             {heading}
--- a/src/Plot.tsx
+++ b/src/Plot.tsx
@@ -4,7 +4,7 @@ import { produce } from 'immer';
 import { CSSProperties, Reducer, useMemo, useReducer } from 'react';
 
 import MarkerDefs from './components/Annotations/MarkerDefs';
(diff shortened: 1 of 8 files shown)


src/Plot.tsx
```diff
@@ -130,6 +134,29 @@ export default function Plot(props: PlotProps) {
               </g>
 
               {axes}
+
+              <Tracking
+                onClick={(position) => {
```
stropitek, line 139: Could pass undefined if callback is undefined

src/Plot.tsx
```diff
@@ -130,6 +134,29 @@ export default function Plot(props: PlotProps) {
...
+                  onClick?.(
+                    closestCalculation(
+                      position,
+                      closest,
+                      state.series,
+                      axisContext,
+                    ),
+                  );
+                }}
+                onMouseMove={(position) => {
```
stropitek, line 149: ditto

stropitek (commented):
I think you should add the `onMouseEnter` and `onMouseLeave` callbacks to toggle the display of the annotation when the mouse leaves the plot

src/Plot.tsx
targos: More importantly, I think we should not render the <Tracking component at all if there are no callbacks, to fully optimize the common case.

src/types.ts
```diff
@@ -83,6 +83,18 @@ export interface PlotProps {
...
+   */
+  onMouseMove?: (result: TrackingResult) => void;
+  /**
+   * Track values on mouse click
+   */
+  onClick?: (result: TrackingResult) => void;
+  /**
+   * Track closest values
+   */
+  closest?: { x?: boolean; y?: boolean; euclidean?: boolean };
```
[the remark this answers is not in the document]
stropitek: https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMIBsD2BnC2yqYgzADmyA3gLABQy9yEArgusACYRwgD8AXMgBGmTOi4gA3LQbIAHvyEix3KXQYBPBcNHjVAX1q1QkWIhSFMUdqDiRslaQ1kCQTALaDoqmepfvPUPqGdGDqAA7mWLj4AGpw6Ex4ADwAKoyykCDs9hg4eAREJKQAfMgAvA5q9MysHOICKQDaAOQ1bJzczQC66ZnZyGBQicg8aCJWNnbIAkxZECQgEOzeTg0tst29EFn2g8OjFhMgtnjTyLOcC0sr9L7ITc3qmxAZ2-17KAfj1sdTM3NXZa0Aw0WgwWYIMDAIhjNxhIjbMCpLY7NBRfKEYhkYoACjCUEwYWwAmoVWQrDy+AaN2QRAwwAQAGsBDiAG7xRICXLRMBxBLJFLFACU5VKrMwHH0ItJMigEDATCgIGQrnQ6CCoJohDhCPAOJlDApPJJjhk1RY7XqAyGEAANKbkHp7WS6WwmSz2egRWVSgazcgAPQB5CXUBLB0yBBEfCMC11bjlZCegB0bXjkgjDCDKswYBD8zD7Ez9CjIBjskTKdkNJBeiFwSAA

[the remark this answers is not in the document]
targos: What if we need multiple closest values at the same time?

stories/control/tracking.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  return (
+    <div>
+      <Plot
+        {...plot}
+        margin={{ bottom: 45, left: 90, top: 40, right: 20 }}
+        onMouseMove={({ coordinates, event: { clientX, clientY } }) =>
+          setHover({ coordinates, position: { x: clientX, y: clientY } })
+        }
+        onClick={({ series }) => setClosest(series)}
```
targos, line 46: result => result.getClosest('euclidean')

src/types.ts
```diff
@@ -175,6 +195,25 @@ export interface MarkersProps {
...
+  point: SeriesPointType;
+  axis: AxisContextType[];
+}
+export interface TrackingResult {
+  event: React.MouseEvent<SVGRectElement, MouseEvent>;
+  coordinates: Record<string, number>;
+  series?: Record<
+    string,
+    { closestX?: ClosestInfo; closestY?: ClosestInfo; closest?: ClosestInfo }
+  >;
```
targos, line 208:
```suggestion
  getClosest('x' | 'y' | 'euclidean'): Record<string, ClosestInfo>;
```

targos (approved):

PR #175, opened 2021-04-14
chore: add mass example

lpatiny:


lpatiny:
This PR is related to https://github.com/zakodium/react-plot/issues/74

Are we currently able to add only vertical line ? 

Should this be a scatter plot with line to 0 ?

maasencioh:
@lpatiny do you mean to do something like this? https://zakodium.github.io/react-plot/?path=/story/api-barseries--bar-control

PR #180, opened 2021-04-16
Multiple axis

### LINKED ISSUE #178: Multiple display of an axis
The current way to show the same axis at top and bottom is to do something like
```tsx
<Axis position="top" id="x" />
<Axis position="bottom" id="x" />
```

The problem is that if any of the axes is removed, the other axis is removed from the context. I think of two options:
1. It should be checked if there's another axis with the same id before removing it from the context
2. An axis could accept several positions (non-orthogonal between them)

What do you think? @targos @lpatiny @stropitek @Sebastien-Ahkrin

maasencioh:
closes #178

### DIFF
--- a/src/Plot.tsx
+++ b/src/Plot.tsx
@@ -69,10 +69,7 @@ export default function Plot(props: PlotProps) {
   const plotHeight = height - top - bottom;
 
   // Set scales
-  const axisContext = useAxisContext(state, {
-    plotWidth,
-    plotHeight,
-  });
+  const axisContext = useAxisContext(state, { plotWidth, plotHeight });
 
   const labels = useMemo(
     () => state.series.map(({ id, label }) => ({ id, label })),
--- a/src/PlotObject.tsx
+++ b/src/PlotObject.tsx
@@ -1,5 +1,6 @@
 import Plot from './Plot';
 import Axis from './components/Axis';
+import ParallelAxis from './components/Axis/ParallelAxis';
 import Legend from './components/Legend';
 import LineSeries from './components/LineSeries';
 import ScatterSeries from './components/ScatterSeries';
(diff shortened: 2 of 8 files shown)


targos:
Could you make a new story (or update one that is not in the API section) please?
The stories in API are supposed to document the API of the components. It is confusing to have this "duplicate" control that is unrelated to it.

src/plotReducer.ts
```diff
@@ -18,15 +18,24 @@ export function reducer(state: State, action: ReducerActions) {
       let currentAxis = state.axis[id];
       if (currentAxis) {
         validatePosition(currentAxis.position, position, id);
-        state.axis[id] = { ...currentAxis, position, ...values };
+        state.axis[id] = {
+          ...currentAxis,
+          position,
+          ...values,
```
targos, line 24: So this makes a merge of the configuration of both axes? We can merge this now if you want but I consider that a bug.

maasencioh: This was something that I wasn't sure if we want them separately and it was the current behavior, if this is going to be saved separately then how we decide the min value in a case like this? Should we throw an error?
```tsx
<Axis min={0} id="x" position="top" />
<Axis min={-1} id="x" position="bottom" />
```

targos: It's a good question, and I think we need to answer it anyway, even with the merge behavior.

maasencioh: The merge behavior takes always the last value, so in the end, we could keep doing that for the moment for all the properties that modify the series context

targos: Something we could do is have another component (for example named `<SideAxis>`) which can only define some of the axis properties and can only be rendered if a "main" axis with the same ID exists.

maasencioh: I like that idea, let me think on the API

targos (approved):

PR #217, opened 2022-01-09
docs : Advenced mass example

wadjih-bencheikh18:


targos:
Is `advenced-mass.json` a copy of https://github.com/zakodium/react-plot/blob/main/scripts/mass/HCys100OH_0.01.json ? We should not have the same identical file twice. @lpatiny I don't know if you wanted to keep the json in the `scripts` folder (and we import it from there) or if it should be moved?

lpatiny:
Indeed the file is a copy but we should rather delete the one in script and keep the data in the 'stories' (stories/data/advenced-mass.json)

@wadjih-bencheikh18 Please also take care of the labels that is the more difficult part of this example.

Please also add in the story book the possibility to force the from / to of the xAxis so that by entering manually the right values we could have something like

image

targos:
advenced -> advanced

wadjih-bencheikh18:
> Indeed the file is a copy but we should rather delete the one in script and keep the data in the 'stories' (stories/data/advenced-mass.json)
> 
> @wadjih-bencheikh18 Please also take care of the labels that is the more difficult part of this example.
> 
> Please also add in the story book the possibility to force the from / to of the xAxis so that by entering manually the right values we could have something like
> 
> image
> 

Yes ofc i'm working on it
I need another example if it's possible

wadjih-bencheikh18:
i couldn't understand "possibility to force the from / to of the xAxis" well
but i added a simple zoom control to see a part from the plot when the form / is visible

lpatiny:
Problem is not we don't see the code example which is quite annoying in the storybook:

image

Can you display the code ?

Concerning the from / to it is in fact min / max

https://github.com/zakodium/react-plot/blob/main/src/components/Axis/Axis.tsx#L15-L16

By default the axis will set the min and max value depending the series but we are able to force those value. This is the way to zoom.

For the X axis, min and max should be editable.

stories/spectra/mass.stories.tsx
```diff
@@ -43,7 +45,35 @@ export function AdvancedMassExample() {
...
-      <Axis id="x" position="bottom" label="Mass [m/z]" />
+      <Annotations>
+        {advancedData.bestPeaks.map((peak) => (
+          <>
+            <Line
+              key={peak.label}
+              x1={peak.x}
+              x2={peak.x}
+              y1={peak.y}
+              y2={peak.y + 2}
```
lpatiny, line 61: If you put number a properties for x1, x2, y1 and y2 the shift will be based on the values of the axis. This means that in your case if you zoom the labels will be exploded.

This is why we can put either number of strings for the x1, x2, y1, y2 properties. If it is strings the values are in pixels from the (0,0) top left.

This means that we need to create the labels containing:
- the line
- the first text
- the second text
using the "pixels" (a string) to shift the different annotations

And they should be placed in a Groupd and then placed using the value of the axis (meaining x and y are numbers)

https://github.com/zakodium/react-plot/blob/main/src/components/Annotations/Group.tsx

stories/spectra/mass.stories.tsx
```diff
@@ -28,3 +30,69 @@ export function MassExample() {
     </Plot>
   );
 }
+export const AdvancedMassExample = AdvancedMassStory.bind({});
+AdvancedMassExample.args = {
+  zoom: false,
+};
```
targos, line 36: You don't need to use `bind`.
You can add args to the function directly.

stories/spectra/mass.stories.tsx
wadjih-bencheikh18: fixed

PR #224, opened 2022-01-11
feat: update annotations props names

wadjih-bencheikh18:


src/components/Annotations/Ellipse.tsx
```diff
@@ -7,21 +7,21 @@ export interface AnnotationEllipseProps
     SVGProps<SVGEllipseElement>,
     'x1' | 'x2' | 'y1' | 'y2' | 'cx' | 'cy' | 'rx' | 'ry'
```
lpatiny, line 8: I'm wondering if x1, x2, y1 and y2 really exists as props for an ellipse (same is true for a circle) and I don't know how to check this.

src/components/Annotations/Rectangle.tsx
```diff
@@ -1,33 +1,43 @@
 import { SVGProps } from 'react';
 
-import { usePositionAndSize } from '../../hooks';
+import { useRectanglePosition } from '../../hooks';
 
 export interface AnnotationRectangleProps
   extends Omit<
     SVGProps<SVGRectElement>,
     'x1' | 'x2' | 'y1' | 'y2' | 'x' | 'y' | 'width' | 'height'
```
lpatiny, line 8: It seems to me that we should also omit 'stroke' because the use of `color` will be mandatory. Same is true for all the other elements in which we are using `color` instead of `fill` or `stroke`.  @targos WDYT ?

src/components/Annotations/Ellipse.tsx
wadjih-bencheikh18: https://developer.mozilla.org/en-US/docs/Web/API/SVGCircleElement
https://developer.mozilla.org/en-US/docs/Web/API/SVGEllipseElement

src/components/Annotations/Rectangle.tsx
wadjih-bencheikh18: @lpatiny Do you want me to remove them?

src/components/Annotations/Ellipse.tsx
targos: I omitted x1, x2, etc because they exist in the TypeScript types so I thought it would help the user to remove them from intellisense.

src/components/Annotations/Circle.tsx
```diff
@@ -3,28 +3,26 @@ import { SVGProps } from 'react';
 import { useEllipsePosition } from '../../hooks';
 
 export interface AnnotationCircleProps
-  extends Omit<
-    SVGProps<SVGCircleElement>,
-    'x1' | 'x2' | 'y1' | 'y2' | 'cx' | 'cy' | 'r'
-  > {
-  cx: number | string;
-  cy: number | string;
+  extends Omit<SVGProps<SVGCircleElement>, 'cx' | 'cy' | 'r'> {
```
targos, line 6:
```suggestion
  extends Omit<SVGProps<SVGCircleElement>, 'x1' | 'x2' | 'y1' | 'y2' | 'cx' | 'cy' | 'r' | 'fill'> {
```

src/components/Annotations/Circle.tsx
```diff
@@ -3,28 +3,26 @@ import { SVGProps } from 'react';
...
 
 export function Circle(props: AnnotationCircleProps) {
-  const { cx: oldCx, cy: oldCy, r: oldR, ...otherProps } = props;
+  const { x: oldX, y: oldY, r: oldR, color, ...otherProps } = props;
 
   const {
-    cx,
-    cy,
+    x,
+    y,
```
targos, line 18: can you make the hook still return `cx` and `cy`, as that is the name of the true props

src/components/Annotations/Ellipse.tsx
```diff
@@ -3,25 +3,29 @@ import { SVGProps } from 'react';
 import { useEllipsePosition } from '../../hooks';
 
 export interface AnnotationEllipseProps
-  extends Omit<
-    SVGProps<SVGEllipseElement>,
-    'x1' | 'x2' | 'y1' | 'y2' | 'cx' | 'cy' | 'rx' | 'ry'
-  > {
-  cx: number | string;
-  cy: number | string;
+  extends Omit<SVGProps<SVGEllipseElement>, 'cx' | 'cy' | 'rx' | 'ry'> {
```
targos, line 6:
```suggestion
  extends Omit<SVGProps<SVGEllipseElement>, 'x1' | 'x2' | 'y1' | 'y2' | 'cx' | 'cy' | 'rx' | 'ry' | 'fill'> {
```

src/components/Annotations/Line.tsx
```diff
@@ -3,11 +3,12 @@ import { SVGProps } from 'react';
 import { usePosition } from '../../hooks';
 
 export interface AnnotationLineProps
-  extends Omit<SVGProps<SVGLineElement>, 'x1' | 'x2' | 'y2' | 'y2'> {
+  extends Omit<SVGProps<SVGLineElement>, 'x1' | 'x2' | 'y1' | 'y2'> {
```
targos, line 6:
```suggestion
  extends Omit<SVGProps<SVGLineElement>, 'x1' | 'x2' | 'y1' | 'y2' | 'stroke'> {
```

src/components/Annotations/Rectangle.tsx
```diff
@@ -1,33 +1,40 @@
...
 export interface AnnotationRectangleProps
-  extends Omit<
-    SVGProps<SVGRectElement>,
-    'x1' | 'x2' | 'y1' | 'y2' | 'x' | 'y' | 'width' | 'height'
-  > {
-  x: number | string;
-  y: number | string;
-  width: number | string;
-  height: number | string;
+  extends Omit<SVGProps<SVGRectElement>, 'x' | 'y' | 'width' | 'height'> {
```
targos, line 6:
```suggestion
  extends Omit<SVGProps<SVGRectElement>, 'x' | 'y' | 'width' | 'height' | 'fill'> {
```

src/components/Annotations/Shape.tsx
```diff
@@ -57,7 +79,7 @@ export function Shape(props: AnnotationShapeProps) {
 
   return (
     <g transform={`translate(${x}, ${y})`}>
-      <Figure size={props.size} style={props.style} />
+      <Figure size={props.size} style={props.style} color={props.color} />
```
targos, line 82: you could do `style={{ fill: props.color, ...props.style }}` here and remove the changes from the shapes above.

src/components/Annotations/Text.tsx
```diff
@@ -9,16 +9,17 @@ export interface AnnotationTextProps
   > {
```
targos, line 9: Add `| 'fill'` to the line above

src/components/Annotations/Ellipse.tsx
lpatiny: Ok then indeed we need to leave the code like it is now. Thanks.

wadjih-bencheikh18: props omit fixed

src/components/Annotations/Circle.tsx
wadjih-bencheikh18: okey fixed

src/components/Annotations/Shape.tsx
wadjih-bencheikh18: fixed

PR #225, opened 2022-01-11
feat : Create Polyline & Polygon Annotation

wadjih-bencheikh18:


lpatiny:
Could you also implement polygon (even if probably there is shared code that should be in a common function) and a corresponding example.

For now I would remove the possibility to display the points because it will yield to many problems like style or shape customisation and I currently don't see any application for it. The goal is not to reproduce a series but just annotate a chart. If ever we really wanted points in the corner of the polyline we can still add Shape annotations.

wadjih-bencheikh18:
Do you mean by "implement polygon" creating new story or new annotation?

lpatiny:
> Do you mean by "implement polygon" creating new story or new annotation?

Both.

Actually for 'polygon' I would say that the property `color` is the fill color while for the annotation `polyline` the property `color` is the stroke color

wadjih-bencheikh18:
polygon added
the color prop is already in polygon 
but in polyline I used line annotation so the prop "color" will be added when #224 merged

src/hooks.ts
```diff
@@ -73,3 +73,17 @@ function convertValueAbs(
   if (scale === undefined) return 0;
   return typeof value === 'number' ? Math.abs(scale(0) - scale(value)) : value;
 }
+
+export function usePointPosition(config: UsePositionConfig[]) {
+  const { axisContext } = usePlotContext();
+  const [xScale, yScale] = validateAxis(axisContext, 'x', 'y');
+  const points = config;
+  let result = '';
+  points.forEach((point) => {
```
lpatiny, line 82: I didn't benchmark this code but I don't think it is the best way to do it because you create huge amount of intermediate strings.

Use rather something like:
```
return points.map((point) => `${convertValue(point.x, xScale)},${convertValue(
      point.y,
      yScale,
    )`).join(' ')
```

wadjih-bencheikh18: yes better

src/components/Annotations/Polyline.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export interface AnnotationPolylineProps
+  extends Omit<SVGProps<SVGLineElement>, 'points'> {
+  points: { x: number | string; y: number | string }[];
+}
+
+export function Polyline(props: AnnotationPolylineProps) {
+  const { points, ...polylineProps } = props;
+  return (
+    <>
```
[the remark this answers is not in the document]
wadjih-bencheikh18: okey no problem

targos:
The new components should be exported in https://github.com/zakodium/react-plot/blob/main/src/components/Annotations/Annotation.tsx

wadjih-bencheikh18:
@targos do i have to open new PR ?

targos:
Yes please

PR #228, opened 2022-01-12
docs : Pca example

wadjih-bencheikh18:
creating pca example using ScatterSeries and ellipse annotation

### DIFF
--- a/stories/examples/pca.stories.tsx
+++ b/stories/examples/pca.stories.tsx
@@ -60,7 +60,9 @@ export function PCAExample() {
                 y="0"
                 style={{
                   fill: fillColor,
-                  transform: `rotate(${-angle}deg)`,
+                  transform: `rotate(${
+                    (angle >= 90 || angle < -90 ? 180 : 0) - angle
+                  }deg)`,
                 }}
               >
                 {label}
--- a/stories/examples/pca.stories.tsx
+++ b/stories/examples/pca.stories.tsx
@@ -2,7 +2,6 @@ import { Meta } from '@storybook/react';
 
 import { Axis, Plot, Heading, ScatterSeries, Annotations } from '../../src';
 import { Ellipse } from '../../src/components/Annotations/Ellipse';
-import { Group } from '../../src/components/Annotations/Group';
 import { Text } from '../../src/components/Annotations/Text';
 import data from '../data/pca.json';
 import { DEFAULT_PLOT_CONFIG } from '../utils';
@@ -37,37 +36,30 @@ export function PCAExample() {
       <Annotations>
         {data.ellipses.map(
           ({ x, y, rx, ry, fillColor, fillOpacity, label, angle }) => (
-            <Group
+            <g
               key={`${x}-${y}`}
-              x={x}
-              y={y}
-              horizontalAlign="middle"
-              verticalAlign="middle"
+              style={{
+                transform: `rotate(${
+                  (angle >= 90 || angle < -90 ? 180 : 0) - angle
+                }deg)`,
+                transformOrigin: 'center',
+                transformBox: 'fill-box',
+              }}
             >
               <Ellipse
                 rx={rx}
                 ry={ry}
-                x="0"
-                y="0"
+                x={x}
+                y={y}
+                color={fillColor}
                 style={{
-                  fill: fillColor,
                   opacity: fillOpacity,
-                  transform: `rotate(${-angle}deg)`,
                 }}
               />
-              <Text
-                x="0"
-                y="0"
-                style={{
-                  fill: fillColor,
-                  transform: `rotate(${
-                    (angle >= 90 || angle < -90 ? 180 : 0) - angle
-                  }deg)`,
-                }}
-              >
+              <Text x={x} y={y} color={fillColor}>
                 {label}
               </Text>
-            </Group>
+            </g>
           ),
         )}
       </Annotations>
--- a/stories/examples/pca.stories.tsx
+++ b/stories/examples/pca.stories.tsx
@@ -3,6 +3,7 @@ import { Meta } from '@storybook/react';
 import { Axis, Plot, Heading, ScatterSeries, Annotations } from '../../src';
 import { Ellipse } from '../../src/components/Annotations/Ellipse';
 import { Group } from '../../src/components/Annotations/Group';
+import { Line } from '../../src/components/Annotations/Line';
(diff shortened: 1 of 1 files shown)


stories/examples/pca.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <Annotations>
+        {data.ellipses.map(
+          ({ x, y, rx, ry, fillColor, fillOpacity, label, angle }) => (
+            <Group
+              key={`${x}-${y}`}
+              x={x}
+              y={y}
+              horizontalAlign="middle"
+              verticalAlign="middle"
+            >
```
lpatiny, line 46: Does it work without group ? I would just create ellipse and text elements with the corresponding `x`, `y` values. I'm afraid that if we have a big text by using horizontal / vertical align we will move the center of the ellipse and this should never happen.

targos: I would also implement it without a group

wadjih-bencheikh18: it doesn't work without group
image

wadjih-bencheikh18: the problem was in the rotation

targos: I see. Maybe we should add a 4th option to the align props of the Annotation groups: `none`, which would not render a GroupAlign, but just a `g` ?

wadjih-bencheikh18: In this case , when i don't use group the transfer origin is the same as the plot (this is why the rotation is not done correctly)

lpatiny: Well this is the reason this example was created. We knew that there could be problem with rotation.

I think when we rotate we always want to do it using the center of the shape or group as origin. I wonder if this could help: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform-origin (transform-origin: center)

wadjih-bencheikh18: Yes i already used transform origin in te style but we still have the same problem
image

lpatiny: Seems adding ` transform-box: fill-box;  ` could help

I did this testcase

```svg
<svg height="500" width="500">
  <ellipse cx="100" cy="100" rx="100" ry="20" style="fill:yellow" />
  <ellipse cx="200" cy="200" rx="100" ry="20"
   style="fill:white; transform-box: fill-box; transform-origin: center; transform:rotate(45deg);"/>
  <ellipse cx="200" cy="200" rx="100" ry="20" style="fill:white; "/>
</svg>
```

image

targos: I don't think we need such tricks. It works fine with a simple `g` (the point of reference becomes the position of the group):

image

targos: I only changed this to test my hypothesis:
image

wadjih-bencheikh18: yes it's working thank you

wadjih-bencheikh18: > I only changed this to test my hypothesis: image

Do you want me to use it?

wadjih-bencheikh18: > I see. Maybe we should add a 4th option to the align props of the Annotation groups: `none`, which would not render a GroupAlign, but just a `g` (it could be the default behavior) ?

So we can add this option to Group annotation

targos: > > I see. Maybe we should add a 4th option to the align props of the Annotation groups: `none`, which would not render a GroupAlign, but just a `g` (it could be the default behavior) ?
> 
> So we can add this option to Group annotation

I created https://github.com/zakodium/react-d3-utils/issues/16 for this.

stories/examples/pca.stories.tsx
```diff
@@ -0,0 +1,82 @@
+import { Meta } from '@storybook/react';
+
+import { Axis, Plot, Heading, ScatterSeries, Annotations } from '../../src';
+import { Ellipse } from '../../src/components/Annotations/Ellipse';
+import { Text } from '../../src/components/Annotations/Text';
+import data from '../data/pca.json';
+import { DEFAULT_PLOT_CONFIG } from '../utils';
+
+export default {
+  title: 'Examples/PCA example',
```
targos, line 10:
```suggestion
  title: 'Examples/PCA of ecstasy',
```

stories/examples/pca.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+    shape: string;
+    fill: string;
+    r: number;
+  };
+  id: string;
+}
+export function PCAExample() {
+  return (
+    <Plot {...DEFAULT_PLOT_CONFIG}>
+      <Heading title="COVID-19 cases in USA (2020)" />
```
targos, line 25:
```suggestion
      <Heading title="Principal component analysis of XTC infrared spectra" />
```

stories/examples/pca.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+      <ScatterSeries
+        markerStyle={{
+          fill: (point: Point) => point.color.fill,
+          stroke: 'none',
```
targos, line 29: Is this needed?

stories/examples/pca.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+                {label}
+              </Text>
+            </g>
+          ),
+        )}
+      </Annotations>
+      <Axis
+        id="x"
+        position="bottom"
+        label="Week"
```
targos, line 69:
```suggestion
        label="PC 1"
```

stories/examples/pca.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+        paddingEnd={0.1}
+        paddingStart={0.1}
+      />
+      <Axis
+        id="y"
+        position="left"
+        label="Number of cases"
```
targos, line 76:
```suggestion
        label="PC 2"
```

stories/examples/pca.stories.tsx
wadjih-bencheikh18: Yes because in default points have black strokes

stories/examples/pca.stories.tsx
wadjih-bencheikh18: fixed

PR #231, opened 2022-01-12
feat: Add color prop to annotation arrow 

wadjih-bencheikh18:


src/components/Plot.tsx
```diff
@@ -195,8 +194,6 @@ export function Plot(props: PlotProps) {
             id={svgId}
             className={svgClassName}
           >
-            <MarkerDefs />
```
wadjih-bencheikh18, line 198: @targos is `MarkerDefs` used just in annotation arrow?

targos: I think, yes.

stories/annotations/arrow.stories.tsx
```diff
@@ -17,6 +17,7 @@ export default {
     y2: 33,
     startPoint: 'none',
     endPoint: 'triangle',
+    color: 'red',
```
targos, line 20: Can you try to add another arrow in the same story with a different color? I am not sure that it will work if we define the shapes more than once with the same ids.

src/components/Plot.tsx
wadjih-bencheikh18: perfect

stories/annotations/arrow.stories.tsx
wadjih-bencheikh18: i got a problem here and tried to fix it (u can check the last commit)

wadjih-bencheikh18: now the color depend on the x1,y1.....
i don't know if it's the perfect solution

targos: It's probably good enough for now. We can revisit it later if it becomes an issue.

src/components/Annotations/MarkerDefs.tsx
```diff
@@ -1,37 +1,46 @@
-export default function MarkerDefs() {
+export default function MarkerDefs({
+  color,
+  data,
+}: {
+  color?: string;
+  data: string;
```
targos, line 6: Let's call it `id` instead of `data`

stories/annotations/arrow.stories.tsx
wadjih-bencheikh18: okey perfect thank you

src/components/Annotations/MarkerDefs.tsx
wadjih-bencheikh18: fixed

PR #233, opened 2022-01-12
Docs : Add example with zoom principle 

wadjih-bencheikh18:
close https://github.com/zakodium/react-plot/issues/199

### DIFF
--- a/stories/control/zoom.stories.tsx
+++ b/stories/control/zoom.stories.tsx
@@ -1,0 +1,87 @@
+import { Meta } from '@storybook/react';
+import { useState } from 'react';
+
+import { Axis, LineSeries, Plot, Annotations, SeriesPoint } from '../../src';
+import { Rectangle } from '../../src/components/Annotations/Rectangle';
+import { DEFAULT_PLOT_CONFIG } from '../utils';
+
+export default {
+  title: 'API/Zoom',
+} as Meta;
+
+const data = [
+  { x: 1, y: 10 },
+  { x: 2, y: 5 },
+  { x: 3, y: 3 },
+  { x: 4, y: 5 },
+  { x: 5, y: 10 },
+];
+
+interface Positions {
+  x1: number;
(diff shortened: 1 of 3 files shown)


maasencioh:
If I single click on the plot it scales strangely

GifMaker_20220113101506702

stories/control/zoom.stories.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+interface ZoomProps {
+  data: SeriesPoint[];
+  displayMarker?: boolean;
+}
+function Zoom({ data, displayMarker }: ZoomProps) {
+  const [position, setPosition] = useState<Positions | null>(null);
+  const [Click, setClick] = useState<boolean>(false);
+  const [min, setMin] = useState<number | undefined>(undefined);
+  const [max, setMax] = useState<number | undefined>(undefined);
```
targos, line 35: Please consolidate this to one state object (with the position and the min/max).
Move the "click" information to a `useRef()`, because the value is not used to render the chart, so it doesn't need to trigger a rerender when it changes.

lpatiny:
Probably when there is only a movement under 5px nothing should happen.

Please also implement the 'double click' that resets the zoom (using onDoubleClick event https://reactjs.org/docs/events.html)

stories/control/zoom.stories.tsx
```diff
@@ -0,0 +1,110 @@
+import { Meta } from '@storybook/react';
+import { useRef, useState } from 'react';
+
+import { Axis, LineSeries, Plot, Annotations, SeriesPoint } from '../../src';
+import { Rectangle } from '../../src/components/Annotations/Rectangle';
+import { DEFAULT_PLOT_CONFIG } from '../utils';
+
+export default {
+  title: 'API/Zoom',
```
targos, line 9:
```suggestion
  title: 'Examples/Zoom',
```

PR #244, opened 2022-01-13
feat: add advanced zoom example story

wadjih-bencheikh18:
close https://github.com/zakodium/react-plot/issues/240

### DIFF
--- a/stories/spectra/mass.stories.tsx
+++ b/stories/spectra/mass.stories.tsx
@@ -172,6 +172,8 @@ export function AdvancedMassExample({ mf }: AdvancedMassExampleProps) {
           )}
         </Annotations>
         <Axis
+          paddingEnd={0.1}
+          paddingStart={0.1}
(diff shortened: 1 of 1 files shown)


lpatiny:
Just use 'HCys100OH' as default value for `mf` (instead of HCys10OH). It will be more interesting.

There is still the bestPeaks recalculation on zoom changing missing but otherwise it looks great.

wadjih-bencheikh18:
bestPeaks recalculation done

lpatiny:
Works as expected, thanks. @targos it is the example we spoke about when the 'best' peaks are recalculate on zoom. If you could have a quick look at the implementation would be great.

Here is the direct link to the storybook: https://deploy-preview-244--react-plot.netlify.app/?path=/story/experimental-spectra-mass--advanced-mass-example

targos:
implementation is good

PR #247, opened 2022-01-13
Feat : add strokeWidth & markerWidth props to arrow annotation

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/243


lpatiny:
Seems to me that all the props should be forwarded like for Line:

https://github.com/zakodium/react-plot/blob/main/src/components/Annotations/Line.tsx#L15-L22

targos:
Yes, and for the marker, we can add `markerSize` instead of `markerWidth`.

wadjih-bencheikh18:
fixed

PR #254, opened 2022-01-15
implement all missing axis props & fix Position of tick labels

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/214
closes : https://github.com/zakodium/react-plot/issues/239

### DIFF
--- a/src/components/Axis/HorizontalAxis.tsx
+++ b/src/components/Axis/HorizontalAxis.tsx
@@ -30,35 +30,45 @@ export default function HorizontalAxis(props: AxisRendererProps) {
     tickLabelStyle,
   } = props;
 
-  const isBottom = position === 'bottom';
+  const isBottom =
+    (position === 'bottom' && tickPosition !== 'inner') ||
+    (position === 'top' && tickPosition === 'inner');
 
-  const transform = isBottom ? `translate(0, ${plotHeight})` : undefined;
+  const transform =
+    position === 'bottom' ? `translate(0, ${plotHeight})` : undefined;
 
   const ticks = useBBoxObserver();
-  function getTickPosition(
-    position: 'inner' | 'outer' | 'center',
-    primaryTickLength: number,
-  ) {
-    const y1 = position === 'center' ? (isBottom ? -5 : primaryTickLength) : 0;
-    const y2 =
-      position === 'inner'
-        ? isBottom
-          ? -1 * primaryTickLength
-          : primaryTickLength
-        : !isBottom
-        ? -1 * primaryTickLength
-        : primaryTickLength;
-    return (value: number) => {
-      return {
-        line: {
-          x1: value,
-          x2: value,
-          y1: y1,
-          y2: y2,
-        },
-        text: { x1: value, y1: isBottom ? 16 : -12 },
-      };
+  function getTickPosition(value: number) {
+    const { y1, y2, textPosition } = GetTickY();
+    return {
+      line: {
+        x1: value,
+        x2: value,
+        y1: y1,
+        y2: y2,
+      },
+      text: {
+        x1: value,
+        y1: isBottom ? textPosition : -textPosition,
+      },
     };
+  }
+  function GetTickY() {
+    const y = isBottom ? primaryTickLength : -1 * primaryTickLength;
+    const textPosition = tickLabelStyle?.fontSize
+      ? parseFloat(tickLabelStyle.fontSize.toString())
+      : 11;
+    return tickPosition === 'center'
+      ? {
+          y1: y / 2,
+          y2: -y / 2,
+          textPosition: textPosition + primaryTickLength / 2,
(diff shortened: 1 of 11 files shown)


src/components/Axis/HorizontalAxis.tsx
```diff
@@ -28,38 +35,55 @@ export default function HorizontalAxis(props: AxisRendererProps) {
...
-        x1: value,
-        x2: value,
-        y2: !isBottom ? -5 : 5,
-      },
-      text: { x1: value, y1: isBottom ? 16 : -12 },
+  function getTickPosition(
+    position: 'inner' | 'outer' | 'center',
+    primaryTickLength: number,
+  ) {
+    const y1 = position === 'center' ? (isBottom ? -5 : primaryTickLength) : 0;
```
targos, line 42: There's some confusion in this function:
- when position is `inner`, the tick should be drawn inside the graph (below the top axis or above the bottom axis)
- when position is `outer`, the tick should be drawn outside the graph (above the top axis or below the bottom axis)
- when position is `center`, the tick should be in between (shifted by `primaryTickLength / 2`)

The text coordinates should also depend on the tick length and position of the ticks, instead of being hardcoded to 16/-12

src/components/Axis/VerticalAxis.tsx
```diff
@@ -29,14 +36,34 @@ export default function VerticalAxis(props: AxisRendererProps) {
...
-        y1: value,
-        y2: value,
-        x2: !isRight ? -5 : 5,
-      },
-      text: { y1: value, x1: isRight ? 10 : -10 },
+  function getTickPosition(
+    position: 'inner' | 'outer' | 'center',
+    primaryTickLength: number,
+  ) {
+    const x1 =
```
targos, line 43: Same comment as above regarding tick position and length.

src/components/Axis/HorizontalAxis.tsx
wadjih-bencheikh18: fixed

wadjih-bencheikh18:
in src/components/Axis/VerticalAxis.tsx
i used `textPosition: 3 + primaryTickLength,` because we need a margin between tick line and the text (3 is the margin)

wadjih-bencheikh18:
is tickLabel have to be in the plot when tickposition is "inner"?

targos:
> is tickLabel have to be in the plot when tickposition is "inner"?

I think the label should never be inside the plot.
/cc @stropitek @lpatiny

stropitek:
> I think the label should never be inside the plot.

I agree

wadjih-bencheikh18:
fixed

src/components/Axis/HorizontalAxis.tsx
```diff
@@ -28,38 +35,75 @@ export default function HorizontalAxis(props: AxisRendererProps) {
...
+      },
+      text: {
+        x1: value,
+        y1: isBottom ? textPosition : -textPosition,
       },
-      text: { x1: value, y1: isBottom ? 16 : -12 },
     };
   }
-
+  function GetTickY() {
```
targos, line 53: camelCase please

src/components/Axis/VerticalAxis.tsx
```diff
@@ -30,36 +37,70 @@ export default function VerticalAxis(props: AxisRendererProps) {
...
-        x2: !isRight ? -5 : 5,
+        x1: x1,
+        x2: x2,
       },
-      text: { y1: value, x1: isRight ? 10 : -10 },
+      text: { y1: value, x1: isRight ? textPosition : -textPosition },
     };
   }
-
+  function GetTickX() {
```
targos, line 51: ditto

src/components/Axis/VerticalAxis.tsx
```diff
@@ -30,36 +37,70 @@ export default function VerticalAxis(props: AxisRendererProps) {
...
+    const x = isRight ? primaryTickLength : -1 * primaryTickLength;
```
targos, line 52:
```suggestion
    const x = isRight ? primaryTickLength : -primaryTickLength;
```

src/components/Axis/VerticalAxis.tsx
```diff
@@ -30,36 +37,70 @@ export default function VerticalAxis(props: AxisRendererProps) {
...
+    switch (tickPosition) {
+      case 'center':
+        return {
+          x1: x / 2,
+          x2: -x / 2,
+          textPosition: 3 + primaryTickLength / 2,
```
targos, line 58: Can you put the "3" in a constant (textOffset ?) and use it everywhere? In the future it will be an option

src/components/Axis/HorizontalAxis.tsx
wadjih-bencheikh18: fixed

PR #255, opened 2022-01-16
feat :  x / y coordinates as "%"

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/237

### DIFF
--- a/src/hooks.ts
+++ b/src/hooks.ts
@@ -58,9 +58,9 @@ export function useEllipsePosition(props: UseEllipsePositionConfig) {
   };
 }
 
-function ConvertString(value: string, total: number) {
+function convertString(value: string, total: number) {
   return value.endsWith('%')
-    ? (Number(value.substring(0, value.length - 1)) * total) / 100
+    ? (Number(value.slice(0, -1)) * total) / 100
     : Number(value);
 }
 function convertValue(
@@ -69,7 +69,7 @@ function convertValue(
   scale?: ScaleLinear<number, number>,
 ) {
   if (scale === undefined) return 0;
-  return typeof value === 'number' ? scale(value) : ConvertString(value, total);
+  return typeof value === 'number' ? scale(value) : convertString(value, total);
 }
 function convertMinValue(
   value1: string | number,
@@ -79,8 +79,8 @@ function convertMinValue(
 ) {
   if (scale === undefined) return 0;
   return Math.min(
-    typeof value2 === 'number' ? scale(value2) : ConvertString(value2, total),
-    typeof value1 === 'number' ? scale(value1) : ConvertString(value1, total),
+    typeof value2 === 'number' ? scale(value2) : convertString(value2, total),
+    typeof value1 === 'number' ? scale(value1) : convertString(value1, total),
   );
 }
 function convertValueAbs(
@@ -91,7 +91,7 @@ function convertValueAbs(
   if (scale === undefined) return 0;
   return typeof value === 'number'
     ? Math.abs(scale(0) - scale(value))
-    : ConvertString(value, total);
+    : convertString(value, total);
 }
 function convertDimensions(
   value1: string | number,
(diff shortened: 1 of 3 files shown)


src/hooks.ts
```diff
@@ -11,47 +11,31 @@ interface UsePositionConfig {
 }
 
 export function usePosition(config: UsePositionConfig) {
-  const { axisContext } = usePlotContext();
+  const { axisContext, width, height } = usePlotContext();
```
targos, line 14: It should use plotWidth and plotHeight. Width and height are the size of the entire SVG

src/hooks.ts
```diff
@@ -62,62 +46,81 @@ interface UseEllipsePositionConfig {
...
-    rx: convertValueAbs(rx, xScale),
-    ry: convertValueAbs(ry, yScale),
+    cx: convertValue(cx, width, xScale),
+    cy: convertValue(cy, height, yScale),
+    rx: convertValueAbs(rx, width, xScale),
+    ry: convertValueAbs(ry, height, yScale),
   };
 }
 
+function ConvertString(value: string, total: number) {
```
stropitek, line 61: our convention is to write regular functions camelCase

stories/annotations/mixed.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+  AnnotationLineProps,
+} from '../../src/components/Annotations/Line';
+
+import { AnnotationPlot } from './annotation.data';
+
+export default {
+  title: 'API/Annotations',
+  component: Line,
+  args: {
+    x1: '40%',
```
stropitek, line 14: Can you maybe add another annotation that uses 0% and 100% so that's it's easier to see that it works properly?

src/hooks.ts
wadjih-bencheikh18: i think there is no need to height and width of the svg

src/hooks.ts
wadjih-bencheikh18: yes okey

src/hooks.ts
wadjih-bencheikh18: after last feat (the rectangle is drawn using x1 x2 y1 y2)

wadjih-bencheikh18: sorry i didn't understand the first time, yes i checked the code and i'm changing it

stories/annotations/mixed.stories.tsx
wadjih-bencheikh18: added

stories/examples/mixed.stories.tsx
```diff
@@ -0,0 +1,29 @@
+import { Meta } from '@storybook/react';
+
+import {
+  Line,
+  AnnotationLineProps,
+} from '../../src/components/Annotations/Line';
+import { AnnotationPlot } from '../annotations/annotation.data';
+
+export default {
+  title: 'Examples/Mixed Anntations',
```
targos, line 10:
```suggestion
  title: 'Examples/Mixed Annotations',
```

wadjih-bencheikh18: fixed

src/hooks.ts
targos: it is not resolved

src/hooks.ts
```diff
@@ -62,62 +46,81 @@ interface UseEllipsePositionConfig {
...
+    cx: convertValue(cx, plotWidth, xScale),
+    cy: convertValue(cy, plotHeight, yScale),
+    rx: convertValueAbs(rx, plotWidth, xScale),
+    ry: convertValueAbs(ry, plotHeight, yScale),
   };
 }
 
+function ConvertString(value: string, total: number) {
+  return value.endsWith('%')
+    ? (Number(value.substring(0, value.length - 1)) * total) / 100
```
targos, line 63:
```suggestion
    ? (Number(value.slice(0, -1)) * total) / 100
```

src/hooks.ts
wadjih-bencheikh18: fixed

PR #256, opened 2022-01-16
docs : Example with annotation callback

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/236


stropitek:
Do you know why it's not possible to see the code in the docs?

targos:
The story should be in Examples, not API

wadjih-bencheikh18:
fixed

PR #261, opened 2022-01-20
Feat : add pan and vertical scale

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/251

### DIFF
--- a/src/components/Tracking.tsx
+++ b/src/components/Tracking.tsx
@@ -21,6 +21,7 @@ export type ClosestInfoResult = Record<string, ClosestInfo<ClosestMethods>>;
 export interface TrackingResult {
   event: React.MouseEvent<SVGRectElement, MouseEvent>;
   coordinates: Record<string, number>;
+  movement?: Record<string, number>;
   getClosest?: (method: ClosestMethods) => ClosestInfoResult;
 }
 export interface KeysResult {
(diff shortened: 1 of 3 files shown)


wadjih-bencheikh18:
there is two problems
1- when we have axis padding the zoom is not perfect (we cannot calculate the plot width ...)
2- plot drag is slow

stories/spectra/mass.stories.tsx
```diff
@@ -72,62 +84,120 @@ export function AdvancedMassExample({ mf }: AdvancedMassExampleProps) {
...
-                x1: position ? position.x1 : x,
-                x2: x,
-              },
-              min: min,
-              max: max,
-            }));
+            if (alt) {
+              setPositions((positions) => ({
+                ...positions,
+                maxX: hand.x - x + maxX,
```
lpatiny, line 168: You should probably use https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/movementX and don't keep any previous position in variables

src/components/Tracking.tsx
```diff
@@ -31,11 +36,55 @@ export interface TrackingProps {
...
+    x: { min: scaleX.invert(0), max: scaleX.invert(plotWidth) },
+    y: { min: scaleY.invert(plotHeight), max: scaleY.invert(0) },
+  };
+}
+function infoFromWheel(
+  event: React.WheelEvent<SVGRectElement>,
+  axisContext: Record<string, PlotAxisContext>,
+  plotHeight: number,
+) {
+  const ratio = 1 + event.deltaY * -0.001;
```
lpatiny, line 66: Sorry my fault. The issue was not clear enough.
Actually the vertical scaling should, in this case, keep the 0 where it is.
So you should always change minY and maxY in order to keep the 0 at the same place and you may ignore the position of the mouse.

lpatiny (changes requested):

src/components/Tracking.tsx
wadjih-bencheikh18: I fixed the y of the mouse while zooming
So the user can zoom where he wants to

stories/spectra/mass.stories.tsx
```diff
@@ -36,23 +36,35 @@ export function MassExample() {
...
-  });
+  const [{ rectangle, hand, minX, maxX, minY, maxY, alt }, setPositions] =
+    useState<Positions | null>({
+      hand: null,
+      rectangle: null,
+      minX: undefined,
+      maxX: undefined,
+      minY: undefined,
+      maxY: undefined,
+      alt: false,
```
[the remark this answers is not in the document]
wadjih-bencheikh18: i'm tracking alt  because there is so many senarios possible to fix 
clicking alt while there's rectangle ...etc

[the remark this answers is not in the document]
wadjih-bencheikh18: i tried to remove alt state but in this case i'm using cursor (string) for cursor css

stories/spectra/mass.stories.tsx
wadjih-bencheikh18: fixed

lpatiny (commented):
All functionalities are working perfectly now. Thanks !

wadjih-bencheikh18:
> All functionalities are working perfectly now. Thanks !

could you please try it with padding (grab functionalty need to initialise max and min )

targos:
It doesn't work for me.

PR #265, opened 2022-01-23
Feat : Directed ellipse

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/260


src/hooks.ts
```diff
@@ -57,7 +64,39 @@ export function useEllipsePosition(props: UseEllipsePositionConfig) {
...
+    ry: convertValueAbs(width, plotHeight, yScale) / 2,
+    rotation:
+      (y1 > y2 ? -1 : 1) *
+      (x1 > x2 ? -1 : 1) *
+      radsToDegs(
+        Math.asin(distanceXY(x1, y1, x1, cy) / distanceXY(x1, y1, cx, cy)),
+      ),
+  };
+}
+function distanceXY(x1: number, y1: number, x2: number, y2: number) {
```
stropitek, line 97: You can use the `ml-euclidean-distance` which is already a dependency in this project

stropitek (commented):
I think the `width` property is not interpreted the right way when it is a number. Right now you just convert it as if it was given as a `y` coordinate. But actually it represents the euclidean distance along the secondary axis. Its calculation depends on the main axis' orientation (so x and y coordinates)

@lpatiny can you confirm that this is how the `width` property should be interpreted?

wadjih-bencheikh18:
> I think the `width` property is not interpreted the right way when it is a number. Right now you just convert it as if it was given as a `y` coordinate. But actually it represents the euclidean distance along the secondary axis. Its calculation depends on the main axis' orientation (so x and y coordinates)
> 
> @lpatiny can you confirm that this is how the `width` property should be interpreted?

Do you mean, width must calculated using the two scales?

stropitek:
> Do you mean width must calculated using the two scales?

Yes you would need both scales for the calculation. But let's wait on Luc's input

lpatiny:
Indeed the use of the parameter `width` was really bad because we never know on which axis we define the width if it is axis units.

What is more rigorous is to speak about major axis and minor axis like in the result of this method that we will use: https://github.com/mljs/directional-distribution/blob/main/src/__tests__/test.ts#L22-L29

I think we should therefore change the name of the properties and either use:
`majorAxis: {x1, x2, y1, y2}` and `minorAxis: {x1, x2, y1, y2}`

or

`x1, y1, x2, y2, x3, y3, x4, y4`

But the main problem may be that during zoom I'm not sure we are keeping an ellipse. Please keep this branch open. I'm trying to figure out this.

image

src/hooks.ts
wadjih-bencheikh18: fixed

wadjih-bencheikh18:
@lpatiny  @stropitek 
i reimplemented the width calculation :
- i made a projection of the width axis on x and y 
- calculate widthX with xScale and widthY with yScale
- calculate the real width based on the two projection

stropitek:
> I think we should therefore change the name of the properties and either use: `majorAxis: {x1, x2, y1, y2}` and `minorAxis: {x1, x2, y1, y2}`
> 
> or
> 
> `x1, y1, x2, y2, x3, y3, x4, y4`

There are some constraints of an ellipse that can be broken by passing all the coordinates of both axes. We can change the api to have better names but we should keep 5 values as input IMO

wadjih-bencheikh18:
> And with for x and y axis we force the `min` to `0` and `max` to `50`.

i didn't understand this , could you please explain?
rectangle zoom feat added

lpatiny:
This example looks good to me. I just updated the data (please pull the new data). Could you set the min / max on both axes to `0` and `50` (force them instead of letting react-plot make the calculation).

wadjih-bencheikh18:
@stropitek fixed

lpatiny:
LGTM
@stropitek could you merge it if the code is ok for you ?

src/hooks.ts
```diff
@@ -57,7 +65,41 @@ export function useEllipsePosition(props: UseEllipsePositionConfig) {
...
+    widthX:
+      (Math.sin(rotation) * convertValueAbs(width, plotHeight, xScale)) / 2,
+    widthY:
+      (Math.cos(rotation) * convertValueAbs(width, plotHeight, yScale)) / 2,
+  };
+  return {
+    cx,
+    cy,
+    rx: euclidean([x1, y1], [x2, y2]) / 2,
+    ry: Math.sqrt(Math.pow(widthX, 2) + Math.pow(widthY, 2)),
```
stropitek, line 99: It's correct, just more readable to use `euclidean`
```suggestion
    ry: euclidean([0, 0], [widthX, widthY]),
    ry: Math.sqrt(Math.pow(widthX, 2) + Math.pow(widthY, 2)),
```

stories/examples/zoom.stories.tsx
```diff
@@ -0,0 +1,279 @@
...
+        onMouseMove={({ coordinates: { x } }) => {
+          if (click.current) {
+            setPositions(({ position }) => ({
+              position: {
+                x1: position ? position.x1 : x,
+                x2: x,
+              },
+              min: min,
+              max: max,
+            }));
```
stropitek, line 74: It makes more sense to update all the values based on what is passed to the update function. Your min max values might not be what you expect here.

```suggestion
            setPositions((positions) => ({
              ...positions,
              position: {
                x1: positions.position ? positions.position.x1 : x,
                x2: x,
              },
            }));
```

stories/examples/zoom.stories.tsx
```diff
@@ -0,0 +1,279 @@
...
+          }
+        }}
+        onMouseLeave={() => {
+          setPositions({
+            position: null,
+            min: min,
+            max: max,
```
stropitek, line 81: better to use an update function

stropitek (commented):
In general, it's safer to use an update function whenever you are reusing your previous state in `setState`

I commented a few but there are quiet a few other in this PR

I tested the behavior and the zoom seems to be slightly wrong. I don't know if it's an issue with the zoom or with the ellipse. If I select right at the borders of the ellipse, I zoom into a larger part

wadjih-bencheikh18:
> In general, it's safer to use an update function whenever you are reusing your previous state in `setState`
> 
> I commented a few but there are quiet a few other in this PR
> 
> I tested the behavior and the zoom seems to be slightly wrong. I don't know if it's an issue with the zoom or with the ellipse. If I select right at the borders of the ellipse, I zoom into a larger part
> 
> 
> 

It's not a issue this behavior is related to padding

targos:
> It's not a issue this behavior is related to padding

Ok, then can you please remove the axis padding in this story? It will be more intuitive

src/hooks.ts
wadjih-bencheikh18: fixed

stories/examples/zoom.stories.tsx
wadjih-bencheikh18: fixed

PR #267, opened 2022-01-23
feat: Axis secondaryTick

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/264

### DIFF
--- a/src/components/Axis/Ticks.tsx
+++ b/src/components/Axis/Ticks.tsx
@@ -1,6 +1,8 @@
 import { CSSProperties, ReactNode, SVGAttributes } from 'react';
 import { PrimaryLinearTicks, PrimaryLogTicks } from 'react-d3-utils';
 
+import { Scales } from './types';
+
 interface CoordinatesXY {
   x1?: number;
   x2?: number;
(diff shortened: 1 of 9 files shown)


src/components/Axis/Ticks.tsx
```diff
@@ -36,52 +40,40 @@ export function Ticks(props: Omit<TicksProps, 'children'>) {
   const {
     primaryTicks,
     getPositions,
-    // hiddenSecondaryTicks = false,
+    secondaryTickLength = 0,
```
targos, line 43: Default value should be 2

targos (changes requested):
There is something wrong with the implementation. If I put `secondaryTickLength={1}`, the length is the same as the primary ticks (which has a length of 5).

image

src/components/Axis/Ticks.tsx
wadjih-bencheikh18: fixed

targos (changes requested):
https://deploy-preview-267--react-plot.netlify.app/?path=/story/api-parallelaxis--control

Parallel axis should have the same default.

targos:
Maybe we should use 3 as the default value? With 2 it seems to short to me.

@lpatiny @stropitek wdyt?

stropitek:
> Maybe we should use 3 as the default value? With 2 it seems to short to me.

Yes I had the same impression when I looked at the stories

wadjih-bencheikh18:
default value changed

targos (approved):
LGTM as a first implementation. We can iterate in other issues/prs if necessary

PR #270, opened 2022-01-24
Feat : Axis Time scale (accept Date type)

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/78

### DIFF
--- a/src/components/Axis/Axis.tsx
+++ b/src/components/Axis/Axis.tsx
@@ -1,4 +1,4 @@
-import { ScaleLinear, ScaleLogarithmic } from 'd3-scale';
+import { ScaleLinear, ScaleLogarithmic, ScaleTime } from 'd3-scale';
 import { CSSProperties, ReactNode, useEffect } from 'react';
 
 import {
@@ -10,6 +10,7 @@ import { getInnerOffset } from '../../utils/axis';
 
 import LinearAxis from './LinearAxis';
 import LogAxis from './LogAxis';
+import TimeAxis from './TimeAxis';
 
 export interface AxisProps {
   id?: string;
@@ -23,7 +24,7 @@ export interface AxisProps {
   paddingEnd?: number;
 
   flip?: boolean;
-  scale?: 'linear' | 'log';
+  scale?: 'linear' | 'log' | 'time';
 
   /**
    * Hide all axis elements.
@@ -167,6 +168,14 @@ export function Axis({
       <LinearAxis
         {...childProps}
         scale={currentAxis.scale as ScaleLinear<number, number>}
+      />
+    );
+  }
+  if (scale === 'time') {
+    return (
+      <TimeAxis
+        {...childProps}
+        scale={currentAxis.scale as ScaleTime<number, number>}
       />
     );
   } else {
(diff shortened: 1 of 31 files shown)


src/components/Axis/Axis.tsx
```diff
@@ -47,6 +48,14 @@ export interface AxisProps {
...
   hiddenTicks?: boolean;
   tickPosition?: TickPosition;
+  /**
+   * LabelsFormat.
+   * the default value:
+   * in linear and log scale
+   * (value)=>String(value)
+   * in time scale
+   * Smart d3 tickFormat
+   */
```
stropitek, line 58:
```suggestion
  /**
   * With time scale the default values is d3's smart tickFormat
   * With other types of scales the default is converting the value to a string
   */
```

stories/control/timeaxis.stories.tsx
```diff
@@ -0,0 +1,93 @@
...
+  { x: new Date(2002, 11, 24, 12, 33), y: 40 },
+  { x: new Date(2002, 11, 24, 13, 33), y: 7 },
+  { x: new Date(2002, 11, 24, 14, 33), y: 15 },
+];
+const dataY = [
+  { x: 3, y: new Date(2002, 11, 24, 10, 33) },
+  { x: 6, y: new Date(2002, 11, 24, 11, 33) },
+  { x: 12, y: new Date(2002, 11, 24, 12, 33) },
+  { x: 14, y: new Date(2002, 11, 24, 13, 33) },
+  { x: 15, y: new Date(2002, 11, 24, 14, 33) },
```
stropitek, line 34: If it's supposed to work with numbers (epochs) as well, can you use numbers on one fothe axis?

src/components/BarSeries.tsx
```diff
@@ -100,7 +100,7 @@ function BarSeriesRender({
       {data.map(({ x, y }) => (
         <line
           style={style}
-          key={`${x}-${y}`}
+          key={`${xScale(x)}-${yScale(y)}`}
```
stropitek, line 103: This change does not seem related to this PR. Does it fix any sort of bug?

wadjih-bencheikh18:
Yes sorry there is many types error to fix

src/components/BarSeries.tsx
targos: I think it's wrong to use x and y as the key. We can replace them with the array index as there is nothing that clearly identifies each data point.

wadjih-bencheikh18: i used index

stories/control/timeaxis.stories.tsx
wadjih-bencheikh18: yes you can use numbers but it must be in milliseconds
check nasdaq example

wadjih-bencheikh18:
types updates

- **coordinates** : number | string | Date
- **max/min in axis** : number (in time scale we use milliseconds)

stropitek:
I think the support of dates complicate the types and the code a lot. We may want to add support to it but maybe at a later point.

It's easy for the user to iterate over the data and convert it to a number before passing it to react-plot.

Can you revert to accepting numbers only?

wadjih-bencheikh18:
> Can you revert to accepting numbers only?

i couldn't revert but I created a new PR

Issue #78, opened 2021-01-15
Time scale


lpatiny:
Specific values for props in Axis
* `scale`=`time` (default value is `linear` and therefore will not render time)

If scale=time we should use the scale from react-d3-utils: https://github.com/zakodium/react-d3-utils/issues/17

@wadjih-bencheikh18 Please could you create a PR in react-d3-utils before working on this issue.

We send the data in 'ms' so dates are simply epoch in `ms`.

lpatiny:
Could we add a callback in the axis component that would format the value ? 

In this case we could have in order to switch the 'epoch' time axis to days a formatter like

`tickLabelCallback: (value) => new Date(value).toISOString().substring(0,10)`

This would allow to improve this example and have readable time series:

https://zakodium.github.io/react-plot/?path=/story/examples-nasdaq--nasdaq-example

PR #272, opened 2022-01-25
Feat : simple time scale (Date type not allowed)

wadjih-bencheikh18:
closes : #78


stropitek (commented):
Is it still necessary to keep the calls to `toNumber` in various places. Shouldn't the only place where it is called be the stories which use data and need to convert to numbers?

wadjih-bencheikh18:
> Is it still necessary to keep the calls to `toNumber` in various places?. Shouldn't the only place where it is called be the stories which use data and need to convert to numbers?

Yes because in react-d3-utils they accept DATE type so scale.invert return DATE |number so we need to use toNumber()

stropitek:
There is a small conflict

wadjih-bencheikh18:
conflict fixed

PR #274, opened 2022-01-26
fix: disable text selection on all plot elements by default

targos:
Closes: https://github.com/zakodium/react-plot/issues/273


lpatiny:
What about the Text Annotation ? https://zakodium.github.io/react-plot/?path=/story/api-annotations--annotation-text. I think it should also be non selectable.

targos:
You need to test with the preview link, not the live site.

targos:
Text annotations are not selectable here: https://deploy-preview-274--react-plot.netlify.app/?path=/story/api-annotations--annotation-text

PR #286, opened 2022-01-29
Feat :Display labels & customize Markers in ScatterSeries (PCA)

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/151


stories/examples/pca.stories.tsx
```diff
@@ -1,76 +1,200 @@
...
+        point1: { x: x1, y: y1 },
+        point2: { x: x2, y: y2 },
+      } = majorAxis;
+      return {
+        x1,
+        y1,
+        x2,
+        y2,
+        ...position,
+        width: rMinor,
```
lpatiny, line 76:
```suggestion
        width: rMinor * 2,
```

lpatiny:
It works as I expected (small mistake with the radius used for width).

lpatiny:
@wadjih-bencheikh18 Please also check the related issue: https://github.com/zakodium/react-plot/issues/297

src/components/ScatterSeries.tsx
```diff
@@ -126,18 +142,20 @@ function ScatterSeriesRender({
...
+      const Marker = markersComps[functionalShape(markerShape, point, i, data)];
+      const label = functionalLabel(pointLabel, point, i, data);
+      const labelStyle = functionalStyle({}, pointLabelStyle, point, i, data);
       return (
         <g // eslint-disable-next-line react/no-array-index-key
           key={`markers-${i}`}
           transform={`translate(${xScale(point.x)}, ${yScale(point.y)})`}
         >
           <Marker size={markerSize} style={{ stroke: style.fill, ...style }} />
+          <text style={labelStyle}>{label}</text>
```
targos, line 158: Maybe we should only render the `text` element if `label` is not empty?

PR #287, opened 2022-01-29
Feat : Function series

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/278


src/components/FunctionSeries.tsx
```diff
@@ -0,0 +1,19 @@
+import { SeriesPoint } from '../types';
+
+import { LineSeries, LineSeriesProps } from './LineSeries';
+
+export interface FunctionSeriesProps extends Omit<LineSeriesProps, 'data'> {
+  data: (x: number) => number;
+  max: number;
+  min: number;
+  step?: number;
```
lpatiny, line 9: The idea of this series is not to have to enter a step and that it is being calculated automatically depending the min / max and the size of the chart (number of point for the width).

Please could you remove this parameter ?

I'm also confused by the min and max props. In the issue the idea was to use the min / max of the xAxis. If there is no other series from which we could extract the min and max value it means that we need to explicitly defined min and max on the axis (and not on the series).

Finally in order to test it is better if this storybook has the square zoom capability.

lpatiny (changes requested):

lpatiny:
In the issue there was also the following suggestion:

`As an example you can use one of the plot of the iris dataset and add 2 FunctionSeries. One is the linear regression and the other one would use as developper dependency https://github.com/mljs/regression-polynomial and would calculate a polynomial regression of degree 3.`

That would be really nice to have and show what can be done and how simply it can be achieved.

stropitek:
What should happen if there is no min / max provided to the axis and the only thing displayed is the `FunctionSerie`? Should we use some arbitrary values like `[0, 1]` or throw?

lpatiny:
> What should happen if there is no min / max provided to the axis and the only thing displayed is the `FunctionSerie`? Should we use some arbitrary values like `[0, 1]` or throw?

It should be the same behaviour that if we today put only an annotation (no series) and we don't specify the min / max on the axis.

It seems that it does not throw but does not give any axis value neither. I just tried

```
function Demo() {
  const { Line } = Annotation;
  return (
    <Plot width={300} height={300}>
      <Annotations>
        <Line
          x1="10"
          y1="10"
          x2="100"
          y2="100"
          color="yellow"
          strokeWidth="4"
        />
      </Annotations>
    </Plot>
  );
}
```
on https://react-plot.zakodium.com

wadjih-bencheikh18:
@lpatiny do you have any idea how to get min max ?
`
    const min = x.scale.invert(0);
    const max = x.scale.invert(plotWidth);
`
i used this but it's not working

targos:
> @lpatiny do you have any idea how to get min max ? `const min = x.scale.invert(0); const max = x.scale.invert(plotWidth);` i used this but it's not working

Why is it not working?

wadjih-bencheikh18:
> Why is it not working?

the problem is in y min/max calculation  (i'm trying to fix it)

src/components/FunctionSeries.tsx
```diff
@@ -1,19 +1,32 @@
...
-  for (let i = min; i < max; i += step) {
-    data.push({ x: i, y: oldData(i) });
+  const { data: oldData, step = 0.1, ...otherProps } = props;
+  const {
+    axisContext: { x },
+    plotWidth,
+  } = usePlotContext();
+  const data = useMemo(() => {
+    const min = x ? x.scale.invert(0) : 0;
+    const max = x ? x.scale.invert(plotWidth) : 1;
```
wadjih-bencheikh18, line 21: when we don't have min/max in x axis the min=0 ,max=1
i tired to do what @lpatiny ask ( max=min=0) but in this case y max and min wont generate automaticly so we will need to set max/min of y and x axis (illogic)

wadjih-bencheikh18:
> Why is it not working?

this solution give min , max without padding so when we have padding in x axis we will have a problem

targos:
There seems to be an infinite loop or something like that in the new story.

src/components/FunctionSeries.tsx
```diff
@@ -0,0 +1,33 @@
+import { useMemo } from 'react';
+
+import { usePlotContext } from '../contexts/plotContext';
+import { SeriesPoint } from '../types';
+import { toNumber } from '../utils';
+
+import { LineSeries, LineSeriesProps } from './LineSeries';
+
+export interface FunctionSeriesProps extends Omit<LineSeriesProps, 'data'> {
+  data: (x: number) => number;
```
targos, line 10: this should be renamed to `getY` too

stories/control/function.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+  component: FunctionSeries,
+  args: {
+    max: 10,
+    min: 0,
+    step: 0.001,
+  },
+} as Meta<FunctionSeriesProps>;
+
+export function Control(args: FunctionSeriesProps) {
+  const getY = (x: number) => 4 * Math.sin(2 * x);
```
targos, line 21: Can you move this out of the Control function so it is not recreated on every render?

Also prefer using a function declaration instead of an arrow when possible:

```
function getY() {
 ...
}
```

stories/control/function.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return (
+    <Plot {...DEFAULT_PLOT_CONFIG}>
+      <Legend position="embedded" />
+      <FunctionSeries
+        data={getY}
+        {...args}
+        xAxis="x"
+        yAxis="y"
+        label="y=3*sin(2*x)"
```
targos, line 30:
```suggestion
        label="y=4*sin(2*x)"
```

src/components/FunctionSeries.tsx
wadjih-bencheikh18: fixed

stories/control/function.stories.tsx
wadjih-bencheikh18: fixed

stories/control/function.stories.tsx
wadjih-bencheikh18: okey thank you

src/components/FunctionSeries.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import { LineSeries, LineSeriesProps } from './LineSeries';
+
+export interface FunctionSeriesProps extends Omit<LineSeriesProps, 'data'> {
+  getY: (x: number) => number;
+  step?: number;
+}
+
+export function FunctionSeries(props: FunctionSeriesProps) {
+  const { getY, step = 0.1, ...otherProps } = props;
```
targos, line 15: @lpatiny do you think it makes sense to have a `step` option ? Anyway I think it should default to `1`, because unless the user zooms inside the page, 1=1 pixel.

src/components/FunctionSeries.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const {
+    axisContext: { x },
```
targos, line 17: this should use the `xAxis` prop (default to "x").

src/components/FunctionSeries.tsx
```diff
@@ -0,0 +1,33 @@
...
+    plotWidth,
+  } = usePlotContext();
+  const min = x ? toNumber(x.scale.invert(0)) : 0;
+  const max = x ? toNumber(x.scale.invert(plotWidth)) : 0;
```
targos, line 21: This should use `plotHeight` if the axis is vertical (position = left or right)

src/components/FunctionSeries.tsx
targos: I'm ok with calculating it, but how? What value other than 1 point per unit (meaning a step of 1) do you expect?

targos: We can compute a better value using `Math.ceil(globalThis.devicePixelRatio || 1)` to take page zoom into account.

src/components/FunctionSeries.tsx
wadjih-bencheikh18: fixed

stories/examples/iris-dataset.stories.tsx
```diff
@@ -129,3 +130,110 @@ export function PCA() {
     </div>
   );
 }
+
+export function PCAFunction() {
```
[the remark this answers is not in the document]
wadjih-bencheikh18: it work,
i'm just adding an issue about the case (x axis max/min are undefined and padding exist)

src/components/FunctionSeries.tsx
wadjih-bencheikh18: step=1 result 
image

src/components/FunctionSeries.tsx
wadjih-bencheikh18: the result when x axis max/min are undefined 
image

src/components/FunctionSeries.tsx
targos: Ok, then I misunderstood what `step` means. What I think it should be:
if the plotWidth is `850` for example, we should have `850 * devicePixelRatio` computed points.

targos: Basically, you need to invert each x using the scale, not only the first one. Otherwise it won't work properly on non-linear scales.

wadjih-bencheikh18: i think 1pixel is not visible anyways so we can fix step to 1
when we zoom it still 1 px

lpatiny:
Could you add in the story book x min / max values?

I would like to be able to test in the storybook what is happening when both x min and x max are undefined.

targos:
@lpatiny the stories in the API section are supposed to only let you control the props of the selected component. Now we can by mistake think that min and max are props of the FunctionSeries

lpatiny:
Indeed for the min / max it may be confusing. Just wanted to try if it was working correctly and actually currently it does not work. Changing min/max has no influence on the chart.

image


@targos What do you suggest ? Remove min/max from here and open another issue to create a more advanced example ?

targos:
>  Remove min/max from here and open another issue to create a more advanced example ?

Yes.

wadjih-bencheikh18:
> Indeed for the min / max it may be confusing. Just wanted to try if it was working correctly and actually currently it does not work. Changing min/max has no influence on the chart.

when you change min and max you have to double click the plot (remember that we have rect zoom )

src/components/FunctionSeries.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const step = 1;
+  const data = useMemo(() => {
+    const data: SeriesPoint[] = [];
+    if (x) {
+      const isHorizontal = x
+        ? x.position === 'top' || x.position === 'bottom'
+        : undefined;
+      const end = isHorizontal ? plotWidth : plotHeight;
+      const scale = x?.scale;
+      for (let i = 0; i <= end; i += step) {
```
lpatiny, line 30: `step` seems useless. You can simply use `i++`

lpatiny (changes requested):

targos:
> > Indeed for the min / max it may be confusing. Just wanted to try if it was working correctly and actually currently it does not work. Changing min/max has no influence on the chart.
> 
> when you change min and max you have to double click the plot (remember that we have rect zoom )
> 
> 

We should not have rect zoom here either. You can move the more complex example to another story

src/components/FunctionSeries.tsx
wadjih-bencheikh18: i used a const so we can change it maybe later

lpatiny (approved):

stories/examples/iris-dataset.stories.tsx
```diff
@@ -4,7 +4,8 @@ import { PCA as MlPCA } from 'ml-pca';
 import LinearRegression from 'ml-regression-simple-linear';
 import { ReactElement } from 'react';
 
-import { LineSeries, Plot, ScatterSeries, SeriesPoint, Axis } from '../../src';
+import { Plot, ScatterSeries, SeriesPoint, Axis } from '../../src';
+import { FunctionSeries } from '../../src/components/FunctionSeries';
```
targos, line 8: The new component should be exported like the others from `src/index.ts`, otherwise it is not possible to use it from another project.
Please also update this line and in the other story to import the component from src.

wadjih-bencheikh18: fixed

targos:
For the future, be careful with the commit message format. There should only be a space after the colon, and no space before it.
It's important because our release tooling doesn't understand the message correctly when it has the wrong format.

Correct: `feat: something`
Incorrect: `feat : something` or `feat :something`

Issue #280, opened 2022-01-28
Callback onClick in Legend Series


lpatiny:
There should be a new property 'onClick' that is available on each line of the legend

![image](https://user-images.githubusercontent.com/1484241/151550893-480bd449-87ae-48ec-a0f5-09fb468a3507.png)

You could improve the following story:

https://zakodium.github.io/react-plot/?path=/story/api-legend--with-two-series

to toggle the stroke width between 1 and 3 pixels. This callback will have to give back the 'id' of the series (and therefore this issue has to be done first: https://github.com/zakodium/react-plot/issues/279).

PR #288, opened 2022-01-29
Feat: add onClick ,style callback in legend

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/279
closes : https://github.com/zakodium/react-plot/issues/280


lpatiny:
Not sure how to test. Did you change as asked in the issue the following example that contains 2 series ? https://deploy-preview-288--react-plot.netlify.app/?path=/story/api-legend--with-two-series because this file does not seem to have change in thie PR.

I think the call back could have 2 parameters: id of the series + event. The goal is to know on which series in the legend you clicked and to be able to change something. In issue #280 the goal is to highlight a specific series (we have 2 series and we can toggle the selection of them).

lpatiny:
Here is an example of the kind of functionality we need to achieve. 



Because the legend use the style of the series it should rather be easy at this level. By changing the style of the series the style of the legend will change at the same time.

If you prefer you can create a new example with 5 lines series and more funny style modifications when selected like in the animated gif example.

lpatiny:
To highlight selected Series you could user stroke-width and also this SVG property: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dasharray

wadjih-bencheikh18:
WithTwoSeries story updated

lpatiny:
The first legend storybook crash

image

The functionality is ok but what is missing is the highlight in the legend. The style of the series is the legend is expected to be the same as the style of the series so when you change the thickness of the line it should be reflected in the legend and this is not the case.

stories/control/legend.stories.tsx
```diff
@@ -51,24 +66,48 @@ type TestProps = LegendProps & { hidden: boolean };
...
-      <LineSeries data={data1} xAxis="x" yAxis="y" label="Label line series" />
-
+      <Legend
+        {...otherProps}
+        onClick={(index) => () => updateHiddenSeries(index)}
+      />
+      <LineSeries
+        data={data1}
+        lineStyle={{
+          strokeWidth: highlightSeries[0] ? '5' : '',
```
lpatiny, line 91: Rather than hardcoding the element of the array we could use a callback.

Please allow lineStyle to receive a callback that is called with an object in which one of the properties (and currently the only one) is `id` (the series id).

lpatiny (changes requested):

src/utils.ts
```diff
@@ -84,6 +89,22 @@ export function functionalStyle<T>(
   return style;
 }
 
+/**
+ * Checks the onClick added to a component and if is a function, gets the resulting value
+ */
+export function functionalOnClick(
```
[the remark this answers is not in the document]
wadjih-bencheikh18: There is two callback types 
Simple callback
`(result:MouseEvent) =>void` when we have one Lineseries
`(Id:number)=>(result:MouseEvent) =>void` if we have many Lineseries

stories/control/legend.stories.tsx
```diff
@@ -51,24 +66,48 @@ type TestProps = LegendProps & { hidden: boolean };
 
 export function WithTwoSeries(props: TestProps) {
   const { hidden, ...otherProps } = props;
+  const [highlightSeries, setHighlightSeries] = useState<boolean[]>([
+    false,
+    false,
+  ]);
+  const updateHiddenSeries = (index: number) => {
```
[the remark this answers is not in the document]
wadjih-bencheikh18: yeah it's better now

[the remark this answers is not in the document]
wadjih-bencheikh18: yes theres event param

stories/control/legend.stories.tsx
```diff
@@ -51,24 +66,48 @@ type TestProps = LegendProps & { hidden: boolean };
 
 export function WithTwoSeries(props: TestProps) {
   const { hidden, ...otherProps } = props;
+  const [highlightSeries, setHighlightSeries] = useState<boolean[]>([
+    false,
+    false,
+  ]);
```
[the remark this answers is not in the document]
wadjih-bencheikh18: yeahh better thanks

src/utils.ts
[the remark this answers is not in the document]
wadjih-bencheikh18: > The first legend storybook crash

could you please check again because i don't have this issue

stories/control/legend.stories.tsx
wadjih-bencheikh18: > Rather than hardcoding the element of the array we could use a callback.
could you please explain this

targos: something like this: `strokeWidth: ({id} => highlightSeries[id] ? '5' : '')`

wadjih-bencheikh18: Thanks, fixed

src/types.ts
```diff
@@ -31,9 +31,18 @@ export type CSSFuncProps<T> = {
...
 };
+export type CSSFuncPropsId = {
+  [key in keyof CSSProperties]:
+    | ((id: string) => CSSProperties[key])
+    | CSSProperties[key];
+};
+export type OnClickProp = (
+  result?: React.MouseEvent<SVGGElement, MouseEvent>,
+) => void;
+export type OnClickFuncProps = ((id?: string) => OnClickProp) | OnClickProp;
```
targos, line 42: I don't see a reason for the `(id?: string) => OnClickProp` version. `onClick` could directly receive two arguments (the id and the event).

src/utils.ts
```diff
@@ -84,6 +90,40 @@ export function functionalStyle<T>(
...
+ * Checks the style added to a component and if is a id function , gets the resulting value
+ */
+export function functionalStyleId(
+  elementStyle: CSSFuncPropsId,
+  id: string,
+): CSSProperties {
+  let style: CSSProperties = {};
+  for (const key in elementStyle) {
+    if (typeof elementStyle[key] === 'function') {
+      style[key] = elementStyle[key](id);
```
targos, line 103: please call it with an object `elementStyle[key]({id})` to be ready in case we add other parameters in the future

stories/control/legend.stories.tsx
```diff
@@ -35,10 +36,25 @@ const data2 = [
...
   return (
     <Plot {...DEFAULT_PLOT_CONFIG}>
-      <Legend {...props} />
-      <LineSeries data={data1} xAxis="x" yAxis="y" label="Label line series" />
+      <Legend
+        {...props}
+        onClick={() => {
+          setHighlight((highlight) => !highlight);
+        }}
+        style={{ fontWeight: highlight ? 'bold' : 'normal' }}
```
targos, line 47: Please rename this to `labelStyle` and pass the style to each label, not to the `<g>` element

src/utils.ts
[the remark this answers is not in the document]
lpatiny: I'm testing the build online in the preview and I get on https://deploy-preview-288--react-plot.netlify.app/?path=/story/api-legend--control a crash even in new anonymous window

image

src/types.ts
wadjih-bencheikh18: yes it would be more simple
done

src/utils.ts
wadjih-bencheikh18: fixed and i used an old function "functionalStyle" used to style scatterSeries points (now we can specifiy T )

stories/control/legend.stories.tsx
wadjih-bencheikh18: done

lpatiny:
@wadjih-bencheikh18 The current version still does not apply the style in the legend. The goal is not that the text in the legend becomes bold but only that the representation of the series takes the same style as the series.

Currently the redline or blue line that is just in front of the legend does not change its width even when selected



I would expect because they use the same styling options as the series that they are updated automatically.

wadjih-bencheikh18:
> the representation of the series takes the same style as the series

i did't know , i though we just need a simple representation that we can change each legend style using id
no problem

src/components/Legend.tsx
```diff
@@ -209,39 +213,56 @@ export function Legend({
...
+  index: number;
+  color: string;
+  style?: CSSProperties;
+}) {
   const x = 0;
-  const y = (config.index + 1) * 16 - x - 7;
+  //hard coding values should be fixed
+  const y = (config.index + 1) * 16 - x - 5;
 
-  return <rect x={x} y={y} width="20" height="3" fill={config.color} />;
```
wadjih-bencheikh18, line 203: i found that line in legend was a hard codded rect so i used a line instead (we should fix y position when label fontSize changes)

src/utils.ts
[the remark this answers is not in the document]
wadjih-bencheikh18: fixed

src/components/Legend.tsx
```diff
@@ -196,39 +213,56 @@ export function Legend({
...
 
-function getLineShape(config: { index: number; color: string }) {
+function getLineShape(config: {
+  index: number;
+  color: string;
+  style?: CSSProperties;
+}) {
   const x = 0;
-  const y = (config.index + 1) * 16 - x - 7;
+  //hard coding values should be fixed
```
targos, line 222:
```suggestion
  // TODO: do not hardcode values
```

lpatiny (approved):

PR #294, opened 2022-02-01
Feat: Show hide series from the legend

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/281


wadjih-bencheikh18:
@lpatiny about the "eye" do you want me to use a local image?

wadjih-bencheikh18:
@lpatiny I don't know why it's working in local storybook but it' doesn't in deploy preview

src/contexts/legendContext.tsx
```diff
@@ -13,7 +13,7 @@ import type { ActionType, Shape } from '../types';
 interface LegendLabelState {
   id: string;
   label?: string;
-
+  visibility?: boolean;
```
targos, line 16: Please rename to `isVisible` (everywhere). The value should not be optional.

src/components/Legend.tsx
```diff
@@ -160,8 +162,21 @@ export function Legend({
...
             <g
+              onClick={(event) => {
+                onClick?.({ event, id: value.id });
+                if (showHide) {
+                  legendDispatch({
+                    type: 'ADD_LEGEND_LABEL',
+                    payload: {
+                      ...value,
+                      visibility: !value.visibility,
+                    },
```
targos, line 173: Please add a new action type called `TOGGLE_VISIBILITY`, which only takes the series id and the reducer can just update the boolean.
This part would be changed to : `legendDispatch({ type: 'TOGGLE_VISIBILITY', payload: { seriesId: value.id } })`

src/components/Legend.tsx
```diff
@@ -187,9 +202,19 @@ export function Legend({
...
             onClick={(event) => {
               onClick?.({ event, id: value.id });
+              if (showHide) {
+                legendDispatch({
+                  type: 'ADD_LEGEND_LABEL',
+                  payload: {
+                    ...value,
+                    visibility: !value.visibility,
+                  },
+                });
```
targos, line 212: ditto

src/components/BarSeries.tsx
```diff
@@ -37,22 +37,26 @@ export function BarSeries(props: BarSeriesProps) {
       hidden: !displayMarker,
     };
   }, [color, displayMarker, figure]);
-
+  const visibility = useMemo(() => {
+    const value = legendState.labels.find((label) => label.id === id);
+    return value ? value.visibility : true;
+  }, [id, legendState.labels]);
```
targos, line 43: This same code is used in all series components. Please refactor to a custom hook. `const isVisible = useIsSeriesVisible(id)`

src/components/BarSeries.tsx
```diff
@@ -37,22 +37,26 @@ export function BarSeries(props: BarSeriesProps) {
...
   useEffect(() => {
     legendDispatch({
       type: 'ADD_LEGEND_LABEL',
       payload: {
         id,
+        visibility,
```
targos, line 49: We do not need to pass `isVisible` here. The default value should be `true` when the label is added.

targos (dismissed):

src/components/BarSeries.tsx
wadjih-bencheikh18: this dispatch is used to update every moment so if i set it to true the TOGGLE_VISIBILITY won't work correctly

src/contexts/legendContext.tsx
wadjih-bencheikh18: fixed

src/components/BarSeries.tsx
targos: Only TOGGLE_VISIBILITY should modify `isVisible` you can keep the current state value when `ADD_LEGEND_LABEL` is handled.

wadjih-bencheikh18: same problem

src/components/LineSeries.tsx
```diff
@@ -29,16 +30,16 @@ export function LineSeries(props: LineSeriesProps) {
     ...otherProps
   } = props;
   const lineStyle = functionalStyle({}, OldLineStyle, { id });
+  const isVisible = useIsSeriesVisible(id);
   useEffect(() => {
```
wadjih-bencheikh18, line 34: @targos you can see here
every moment we're adding and removing legend label
(it's not working now)

src/contexts/legendContext.tsx
```diff
@@ -33,7 +33,8 @@ interface LegendState {
 
 type LegendActions =
   | ActionType<'ADD_LEGEND_LABEL', LegendLabelState>
```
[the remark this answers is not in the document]
wadjih-bencheikh18: added , but still don't work

src/components/LineSeries.tsx
wadjih-bencheikh18: we're changing LegendLabels state

src/contexts/legendContext.tsx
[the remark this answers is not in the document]
wadjih-bencheikh18: fixed

stropitek:
@wadjih-bencheikh18 there are conflicts again 🥲

wadjih-bencheikh18:
> In the control and the showHide stories, the showHide prop control does not seem to have any effect

as i mentioned 
i don't know why it's working in local perfectly but here no

stropitek:
I couldn't understand why the netlify build is different from a local build. I implemented a workaround, it seems like the problem is gone if you don't extend the legend's props with other props.

PR #296, opened 2022-02-01
Plot controller

targos:
Work in progress!

TODO:
- [ ] handle mouse wheel
- [ ] smart handle of double click
- [x] handle mouse out of plot during action (to discuss)

Closes: https://github.com/zakodium/react-plot/issues/285

Modified story code: https://github.com/zakodium/react-plot/blob/hooks/stories/examples/zoom.stories.tsx


targos:
The API will probably change but for now it is like that:

Your code needs to be a child of `PlotController`:

```jsx
function Parent() {
  return <PlotController>
    <Child />
  </PlotController>
}

function Child() {
  const plotControls = usePlotControls();

  function handleZoom({min, max}) {
    plotControls.setAxis('x', {min, max});
  }

  function resetZoom() {
    plotControls.resetAxes(['x']);
  }

  return <Plot>{...}</Plot>
}
```

stories/examples/zoom.stories.tsx
```diff
@@ -240,8 +212,8 @@ export function RectangleZoom() {
             />
           )}
         </Annotations>
-        <Axis min={minX} max={maxX} id="x" position="bottom" label="PC 1" />
-        <Axis min={minY} max={maxY} id="y" position="left" label="PC 2" />
+        <Axis min={0} max={50} position="bottom" label="PC 1" />
+        <Axis min={0} max={50} position="left" label="PC 2" />
```
targos, line 216: Note that these are the default min/max values. If the parent plot controller contains values, they take precedence.
Priority is Plot controller > Axis > Series.

PR #304, opened 2022-02-05
Feat: useCrossHair Hook 

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/303

### DIFF
--- a/src/hooks/useCrossHair.tsx
+++ b/src/hooks/useCrossHair.tsx
@@ -1,8 +1,7 @@
 import { CSSProperties, useState } from 'react';
 
-import { Group } from '../components/Annotations/Group';
 import { AnnotationLineProps, Line } from '../components/Annotations/Line';
-import { Text } from '../components/Annotations/Text';
+import { AnnotationTextProps, Text } from '../components/Annotations/Text';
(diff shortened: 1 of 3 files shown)


src/hooks/useCrossHair.tsx
```diff
@@ -0,0 +1,66 @@
+import { CSSProperties, useState } from 'react';
+
+import { Group } from '../components/Annotations/Group';
+import { AnnotationLineProps, Line } from '../components/Annotations/Line';
+import { Text } from '../components/Annotations/Text';
+import { usePlotEvents } from '../contexts/plotController/plotControllerContext';
+
+export interface UseCrossHairOptions {
+  axisId?: { x: string; y: string };
```
targos, line 9: Use these two props like in the rectangular zoom hook: https://github.com/zakodium/react-plot/blob/d86e2e2f11f290f0ce2b8a4842ce16e6e17811e9/src/hooks/types.ts#L3-L6

wadjih-bencheikh18: fixed

stropitek (commented):
code LGTM

Can you add a little extra padding between the crosshairs and the text so that it looks a little bit better?

PR #306, opened 2022-02-05
Feat : xShift & yShift in Series

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/283
closes : https://github.com/zakodium/react-plot/issues/284


lpatiny:
Functionality works as I expected. @wadjih-bencheikh18 Are you able to try to fill the Series with white background to show this effect in which the spectra in the bottom hide the next spectrum, etc. Like in 
image

src/components/RangeSeries.tsx
```diff
@@ -26,27 +27,47 @@ export function RangeSeries<T extends RangeSeriesPoint>(
...
+    horizontalAxisId: xAxis,
+    verticalAxisId: yAxis,
+    xShift: oldXShift,
+    yShift: oldYShift,
+  });
   // Update plot context with data description
   const dispatch = usePlotDispatchContext();
+  const { xShift: xShiftInverted, yShift: yShiftInverted } = useInvertShift({
+    xShift,
+    yShift,
```
stropitek, line 51: shouldn't the axis be passed here as well?

I expect there to be a bug if the default axis are not used

src/hooks.ts
```diff
@@ -162,3 +170,48 @@ export function usePointPosition(config: UsePositionConfig[]) {
...
+
+interface UseInvertOptions extends DualAxisOptions {
+  xShift?: number;
+  yShift?: number;
+}
+export function useInvertShift(options: UseInvertOptions) {
+  const { axisContext } = usePlotContext();
+  const {
+    horizontalAxisId = 'x',
+    verticalAxisId = 'y',
```
stropitek, line 204: ditto

src/hooks.ts
```diff
@@ -162,3 +170,48 @@ export function usePointPosition(config: UsePositionConfig[]) {
...
 }
+interface UseShiftOptions extends DualAxisOptions {
+  xShift?: number | string;
+  yShift?: number | string;
+}
+export function useShift(options: UseShiftOptions) {
+  const { axisContext, plotWidth, plotHeight } = usePlotContext();
+  const {
+    horizontalAxisId = 'x',
+    verticalAxisId = 'y',
```
stropitek, line 181: Those should be mandatory probably to avoid bugs

The default values are already set by the components which use this hook

stropitek (commented):
Other than the comments I made in the code, we should think about tracking. We probably need to add a property that contains the shifted coordinates.

@lpatiny any thoughts on this?

src/components/RangeSeries.tsx
```diff
@@ -26,27 +27,47 @@ export function RangeSeries<T extends RangeSeriesPoint>(
...
 
     const [y1Min, y1Max] = extent(data, (d) => d.y1);
     const [y2Min, y2Max] = extent(data, (d) => d.y2);
 
-    const x = { min: xMin, max: xMax, axisId: xAxis };
+    const x = { min: xMin, max: xMax, shift: xShiftInverted, axisId: xAxis };
     const y = {
       min: Math.min(y1Min, y2Min),
       max: Math.max(y1Max, y2Max),
+      shift: -yShiftInverted,
```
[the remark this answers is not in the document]
wadjih-bencheikh18: Because in the plot we start calculating px from the top left 
So 
Scale.invert(value)-Scale.invert(0)
Will work in Xshift but not in yshift

But i can put the - in the hook (i think it will be better)

src/utils.ts
```diff
@@ -167,6 +167,9 @@ export function dataConvertDate(
 }
 
 export function toNumber(value: number | Date) {
+  if (typeof value === 'undefined') {
```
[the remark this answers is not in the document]
wadjih-bencheikh18: There is a case where axisContext is still empty (before rendering)

src/components/RangeSeries.tsx
wadjih-bencheikh18: Yes true thank you

src/hooks.ts
```diff
@@ -162,3 +170,48 @@ export function usePointPosition(config: UsePositionConfig[]) {
...
+    yShift = 0,
+  } = options;
+  const [xScale, yScale] = validateAxis(
+    axisContext,
+    horizontalAxisId,
+    verticalAxisId,
+  );
+  return {
+    xShift: toNumber(xScale?.invert(xShift)) - toNumber(xScale?.invert(0)),
+    yShift: toNumber(yScale?.invert(0)) - toNumber(yScale?.invert(yShift)),
```
stropitek, line 215: If you are inverting the sign for `y` in places which use this hook, wouldn't it make more sense to just use the same calculation here as for `xShift` which will also invert the sign?

src/utils.ts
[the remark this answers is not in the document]
wadjih-bencheikh18: the case of undefined was not treated , when we try toNumber(undefined) we will have an error 
so now it returns undefined

src/hooks.ts
wadjih-bencheikh18: problem fixed thank you

src/hooks.ts
wadjih-bencheikh18: problem fixed thank you

src/hooks.ts
wadjih-bencheikh18: maybe we should discuss about this because it's used in other hooks

src/components/BarSeries.tsx
```diff
@@ -15,12 +16,25 @@ export function BarSeries(props: BarSeriesProps) {
   const { colorScaler } = usePlotContext();
   const id = useId(props.id, 'series');
   const { lineStyle = {}, displayMarker = false, ...otherProps } = props;
+  const {
+    xAxis = 'x',
+    yAxis = 'y',
+    xShift: oldXShift = '0',
```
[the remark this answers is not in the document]
wadjih-bencheikh18: fixed

src/hooks.ts
stropitek: Public (exported in the main index file) hooks and components should definitely have this optional so that the usual default case is handled with as few props and options as possible

But does it make sense to make useShift and useShiftInverted public?

If not then we should have those properties mandatory and make them private internal functions

If yes then what is the use case? It seems to me like the only purpose of those hooks is to be used internally but I might be missing something

src/hooks.ts
```diff
@@ -139,6 +139,13 @@ function convertToPx(value: string | number, total: number, scale?: Scales) {
     ? scale(value) - scale(0)
     : convertString(value, total);
 }
+function convertToScale(value: string | number, total: number, scale?: Scales) {
+  if (scale === undefined) return 0;
+  return typeof value === 'number'
+    ? value
+    : toNumber(scale.invert(convertString(value, total))) -
+        toNumber(scale.invert(0));
```
stropitek, line 147: It's nice that you just return the value when it's a number now 👍🏻

src/utils.ts
```diff
@@ -168,7 +168,7 @@ export function dataConvertDate(
 
 export function toNumber(value: number | Date) {
   if (typeof value === 'undefined') {
-    return 0;
+    return value;
```
stropitek, line 171: You will be returning the string "undefined" here

I don't think that is what was intended

wadjih-bencheikh18: i want to return undefined if the value is undefined , do you think 0 better? 
or do you have any other suggestion to avoid bugs here?

src/hooks.ts
wadjih-bencheikh18: yes useShift is an internal private function
I totaly agree with you
Thanks

src/utils.ts
stropitek: my bad! my brain omitted `typeof` when doing the review and I thought you were doing `value === 'undefined'`

I'm used to doing `value === undefined` when I write code myself

PR #309, opened 2022-02-06
feat: add support for nested PlotController contexts

targos:
Closes: https://github.com/zakodium/react-plot/issues/298


stropitek (approved):
SGTM

PR #326, opened 2022-02-08
feat: add hooks for tracking closest points

stropitek:
Closes: https://github.com/zakodium/react-plot/issues/314


lpatiny:
Some brainstorming can be found here: https://hackmd.io/-CYbArNQR_KsMriP0uP4Dg

src/components/Tracking.tsx
```diff
@@ -12,11 +12,15 @@ import { closestPoint, toNumber } from '../utils';
...
+  axis: T extends ClosestMethods.euclidean
     ? Record<'x' | 'y', PlotAxisContext>
     : PlotAxisContext;
 }
-export type ClosestMethods = 'x' | 'y' | 'euclidean';
+export enum ClosestMethods {
+  x = 'x',
+  y = 'y',
+  euclidean = 'euclidean',
+}
```
targos, line 23: For libraries, I tend to prefer using unions instead of enums because they are easier to use (no need to import, autocomplete just by writing `''`, interoperable with values that come from somewhere that doesn't know about `react-plot`)

PR #328, opened 2022-02-08
fix: hook examples

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/324
closes : https://github.com/zakodium/react-plot/issues/322
closes : https://github.com/zakodium/react-plot/issues/317

### DIFF
--- a/src/hooks/useAxisWheelZoom.tsx
+++ b/src/hooks/useAxisWheelZoom.tsx
@@ -35,8 +35,8 @@ export function useAxisWheelZoom(options: UseAxisWheelZoomOptions = {}) {
       }) {
         if (event instanceof WheelEvent) {
           const position = center === 'mouse' ? mousePosition : center;
-
-          const ratio = 1 + event.deltaY * (invert ? -0.001 : 0.001) * factor;
+          const delta = event.deltaY * (invert ? -0.001 : 0.001) * factor;
+          const ratio = delta < 0 ? -1 / (delta - 1) : 1 + delta;
           const min = position - (position - oldMin) * ratio;
           const max = position + (oldMax - position) * ratio;
 


wadjih-bencheikh18:
@targos could you please check onKeyUp and onKeyDown added?

stropitek:
> @targos could you please check onKeyUp and onKeyDown added?

He is on vacation this week.

The alt key is captured in the mouse event. You should be able to implement this with changes to the `useStartMoveEnd` hook for adding support of a pressed metaKey on mouse down. (the mouse event contains this information).

```jsx
  useStartMoveEnd({
    metaKey: {
      alt: true,
      shift: false,
    }
  });
```

The hook would do it's thing only if when the user presses down with the mouse, the alt key is pressed.

For now I think it's ok if the hand cursor does not appear when you press the alt key but you haven't yet started the panning gesture.

wadjih-bencheikh18:
> For now I think it's ok if the hand cursor does not appear when you press the alt key but you haven't yet started the panning gesture.

thanks @stropitek

src/hooks.ts
```diff
@@ -199,16 +200,15 @@ interface AxisId {
...
+      min: overrides[xAxis]?.min,
     },
     [yAxis]: {
-      max: toNumber(yScale?.invert(0)),
-      min: toNumber(yScale?.invert(plotHeight)),
+      max: overrides[yAxis]?.max,
+      min: overrides[yAxis]?.min,
     },
   };
 }
```
wadjih-bencheikh18, line 214: @stropitek could you please help me , why `useMaxMin` is not working

wadjih-bencheikh18: the two versions are not working (when i use scale.invert and also when i use overrides)

stropitek: I would suggest that you start by implementing the pan feature without the alt key and without other hooks for zooming and scrolling. That pan feature should use the existing `useStartMoveEnd` hook

wadjih-bencheikh18:
i added return max, min changed in hooks so that the user can use them 
take as an example mass story we need always min,max to update peaks

src/components/Tracking.tsx
```diff
@@ -37,32 +37,32 @@ export interface TrackingProps {
...
-//   event: React.WheelEvent<SVGRectElement>,
-//   axisContext: Record<string, PlotAxisContext>,
-//   plotHeight: number,
-// ) {
-//   const ratio = 1 + event.deltaY * -0.001;
+function infoFromWheel(
+  event: WheelEvent,
+  axisContext: Record<string, PlotAxisContext>,
+  plotHeight: number,
+) {
```
targos, line 44: I think I'd prefer if we could implement the "mouse wheel to zoom" action using the data returned by `infoFromMouse`. We can add more info to it if necessary, but the API will be simpler if all mouse events generate the same info.

src/components/Tracking.tsx
```diff
@@ -178,7 +178,6 @@ type NativeMouseEventType =
   | 'click'
   | 'dblclick'
   | 'wheel';
-
```
targos, line 181: why did you delete all these spaces?

targos:
I feel like you tried to do too much in the same pull request. We are still not sure about the design of the new hooks and doing things more slowly (not all stories at the same time) would have been better imo.

src/components/Tracking.tsx
wadjih-bencheikh18: We don't have yDelta in mouseEvent 
But yes i will try to fix make it simpler

targos: I see. Then can you try to make a common interface that has the event type as a parameter?
Something like `TrackingResult<EventType extends MouseEvent = MouseEvent>`

wadjih-bencheikh18:
> would have been better imo.

Yes i agree , and i think we should let the user change event callbacks (onClick ...) So they can create any functionality they want ,not just use our hooks => we don't limite react-plot

Also i don't know if it a good idea but i want to share it
Instead of using these hooks maybe pass it as a boolean prop to plot will be more simple , or we can add options prop that we can enter every hook/props we want to use

Options={
"AxisZoom":{rectangleStyle,...},
"ScrollZoom"
}

targos:
> Also i don't know if it a good idea but i want to share it
Instead of using these hooks maybe pass it as a boolean prop to plot will be more simple , or we can add options prop that we can enter every hook/props we want to use

It's a good idea, we discussed it briefly last week. I think when the hooks are stabilized, that we will export a new component that wraps `Plot` with a `PlotController` and uses the hooks internally. It could be named something like `ZoomablePlot`.

targos:
> Yes i agree , and i think we should let the user change event callbacks (onClick ...) So they can create any functionality they want ,not just use our hooks => we don't limite react-plot

With the `usePlotEvents` hook, the user can do anything they want. It's like passing the callbacks directly, but with a slightly different and more composable API.

src/components/Tracking.tsx
wadjih-bencheikh18: sorry i didn't pay attention

src/components/Tracking.tsx
```diff
@@ -37,39 +38,34 @@ export interface TrackingProps {
...
+        ratio > 1
+          ? pos + (total - pos) / ratio
+          : total + (total - pos) * (1 / ratio - 1);
+      newDomain[key] = [
+        toNumber(scale.invert(min)),
+        toNumber(scale.invert(max)),
+      ];
+    }
+    return { event, newDomain };
+  }
```
targos, line 68: All of this logic should happen inside the `useAxisWheelZoom` hook.
I'm pretty sure that you can implement it using the current axis' min/max values directly instead of the raw pixel width and height.
So, instead of a condition that returns a different object when there a wheel event, it should be enough to add a fourth dictionary in the result with the domains (min and max) of all axes: `domains: { [key]: domain }`

src/components/Tracking.tsx
```diff
@@ -18,11 +18,12 @@ export interface ClosestInfo<T extends ClosestMethods> {
...
-  event: MouseEvent;
-  coordinates: Record<string, number>;
-  clampedCoordinates: Record<string, number>;
+export interface TrackingResult<EventType extends MouseEvent = MouseEvent> {
+  event: EventType;
+  coordinates?: Record<string, number>;
+  clampedCoordinates?: Record<string, number>;
   movement?: Record<string, number>;
+  newDomain?: Record<string, [number, number]>;
   getClosest?: (method: ClosestMethods) => ClosestInfoResult;
```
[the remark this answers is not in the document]
wadjih-bencheikh18: okey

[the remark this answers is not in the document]
wadjih-bencheikh18: do newDomain must be mandatory ?

src/components/Tracking.tsx
```diff
@@ -18,11 +18,12 @@ export interface ClosestInfo<T extends ClosestMethods> {
...
-export interface TrackingResult {
-  event: MouseEvent;
-  coordinates: Record<string, number>;
-  clampedCoordinates: Record<string, number>;
+export interface TrackingResult<EventType extends MouseEvent = MouseEvent> {
+  event: EventType;
+  coordinates?: Record<string, number>;
+  clampedCoordinates?: Record<string, number>;
   movement?: Record<string, number>;
+  newDomain?: Record<string, [number, number]>;
```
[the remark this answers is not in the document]
wadjih-bencheikh18: okey

stories/spectra/mass.stories.tsx
```diff
@@ -35,32 +43,12 @@ export function MassExample() {
...
-    useState<Positions | null>({
-      rectangle: null,
-      minX: undefined,
-      maxX: undefined,
-      minY: undefined,
-      maxY: undefined,
-      alt: false,
-    });
-  let click = useRef<boolean>(false);
+  const { annotations, max, min } = useAxisZoom();
```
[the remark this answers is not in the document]
wadjih-bencheikh18: i used 

```
interface UsePlotControllerAxisOptions extends ControllerHookOptions {
  axisId?: string;
}
```

src/hooks/useRectangularZoom.tsx
```diff
@@ -13,11 +16,19 @@ export interface UseRectangularZoomOptions
 export function useRectangularZoom(options: UseRectangularZoomOptions = {}) {
   const { horizontalAxisId = 'x', verticalAxisId = 'y' } = options;
 
+  const {
+    axes: { [horizontalAxisId]: oldX, [verticalAxisId]: oldY },
+  } = usePlotOverrides();
+  const [xY, setXY] = useState({
+    [horizontalAxisId]: oldX,
+    [verticalAxisId]: oldY,
+  });
```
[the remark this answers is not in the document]
wadjih-bencheikh18: the usePlotControllerAxis fixed the problem

src/components/Tracking.tsx
wadjih-bencheikh18: done

targos:
The FunctionSeries story is missing the plot controller decorator: https://deploy-preview-328--react-plot.netlify.app/?path=/story/api-functionseries--control

src/hooks/useAxisWheelZoom.tsx
```diff
@@ -0,0 +1,55 @@
...
+        if (event instanceof WheelEvent) {
+          const position = zoomCenter === 'mouse' ? mousePosition : zoomCenter;
+          const ratio = 1 + event.deltaY * -0.001;
+          const min = position - (position - oldMin) * ratio;
+          const max = position + (oldMax - position) * ratio;
+          // const min = ratio > 1 ? pos * (1 - 1 / ratio) : pos * (1 - 1 / ratio);
+          // const max =
+          //   ratio > 1
+          //     ? pos + (total - pos) / ratio
+          //     : total + (total - pos) * (1 / ratio - 1);
```
targos, line 41: Any reason to keep this commented code?

wadjih-bencheikh18: fixed

src/components/Tracking.tsx
```diff
@@ -37,51 +38,29 @@ export interface TrackingProps {
...
+  const domains: TrackingResult['domains'] = {};
   const movement: TrackingResult['movement'] = {};
   for (const key in axisContext) {
-    const { scale, clampInDomain, position } = axisContext[key];
-    if (HORIZONTAL.includes(position)) {
+    const { scale, clampInDomain, position, domain } = axisContext[key];
+    const isHorizontal = HORIZONTAL.includes(position);
+    if (onWheel) {
+      domains[key] = domain;
+    }
```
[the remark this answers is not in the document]
wadjih-bencheikh18: done

wadjih-bencheikh18:
after reading docs
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/movementX
i found that movementX /Y don't give us the exact value so we can see that the movment is not perfect

src/hooks/useAxisWheelZoom.tsx
```diff
@@ -0,0 +1,54 @@
...
+  usePlotControls,
+  usePlotEvents,
+} from '../contexts/plotController/plotControllerContext';
+
+import { ControllerHookOptions } from './types';
+
+export interface UseAxisWheelZoomOptions extends ControllerHookOptions {
+  direction?: 'horizontal' | 'vertical';
+  axisId?: string;
+  zoomCenter?: 'mouse' | number;
```
lpatiny, line 11: just call it `center`

src/hooks/useAxisWheelZoom.tsx
```diff
@@ -0,0 +1,54 @@
...
+  zoomSpeed?: number;
```
lpatiny, line 12: just call it `factor`

src/hooks/index.ts
```diff
@@ -3,3 +3,6 @@ export * from './useDrawRectangle';
 export * from './useRectangularZoom';
 export * from './useStartMoveEnd';
 export * from './useCrossHair';
+export * from './useAxisWheelZoom';
+export * from './usePlotControllerAxis';
+export * from './useGrab';
```
lpatiny, line 8: Please rename `usePan`

src/hooks/useAxisWheelZoom.tsx
```diff
@@ -0,0 +1,54 @@
...
+        event,
+        coordinates: { [axisId]: mousePosition },
+        domains: {
+          [axisId]: [oldMin, oldMax],
+        },
+      }) {
+        if (event instanceof WheelEvent) {
+          const position = zoomCenter === 'mouse' ? mousePosition : zoomCenter;
+
+          const ratio = 1 + event.deltaY * -0.001 * zoomSpeed;
```
lpatiny, line 37:
```suggestion
          const ratio = 1 + event.deltaY * (invert ? -0.001 : 0.001) * zoomSpeed;
```

src/hooks/useAxisWheelZoom.tsx
```diff
@@ -0,0 +1,54 @@
...
+} from '../contexts/plotController/plotControllerContext';
+
+import { ControllerHookOptions } from './types';
+
+export interface UseAxisWheelZoomOptions extends ControllerHookOptions {
+  direction?: 'horizontal' | 'vertical';
+  axisId?: string;
+  zoomCenter?: 'mouse' | number;
+  zoomSpeed?: number;
+}
```
lpatiny, line 13: We should add one more option `invert` by default 'false'. This allows to invert the direction of the scroll. The default behaviour should be opposite to the current one. Meaning when you scroll up the spectrum increase of size.

lpatiny (changes requested):

src/hooks/useAxisWheelZoom.tsx
wadjih-bencheikh18: fixed

src/hooks/useAxisWheelZoom.tsx
wadjih-bencheikh18: prop added

lpatiny:
@wadjih-bencheikh18 Please go on this page: https://deploy-preview-328--react-plot.netlify.app/?path=/story/experimental-spectra-mass--advanced-mass-example and quickly use the scroll wheel up and down.

I still get the crash:

image

PR #332, opened 2022-02-11
feat: add keyboard events

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/331

### DIFF
--- a/src/components/Tracking.tsx
+++ b/src/components/Tracking.tsx
@@ -20,7 +20,7 @@ export interface ClosestInfo<T extends ClosestMethods> {
 export type ClosestMethods = 'x' | 'y' | 'euclidean';
 export type ClosestInfoResult = Record<string, ClosestInfo<ClosestMethods>>;
 export interface TrackingResult {
-  event: MouseEvent | KeyboardEvent;
+  event: MouseEvent;
   coordinates: Record<string, number>;
   clampedCoordinates: Record<string, number>;
   movement?: Record<string, number>;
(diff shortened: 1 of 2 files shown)


targos:
I'm really not convinced we should support keyboard events inside plots. It is not intuitive to need to have focus on the plot for them to work. What kind of example would you implement with this feature?

wadjih-bencheikh18:
> I'm really not convinced we should support keyboard events inside plots. It is not intuitive to need to have focus on the plot for them to work. What kind of example would you implement with this feature?

In the old version we used it to implement grab functionality (click alt btn)
We need it in the initialisation of max/min also to change cursor style

wadjih-bencheikh18:
> > 
> 
> We don't need it to initialize max/min because the mouse event contains the information of which modifier keys are currently pressed.


I thought mouse event only contains coordinates
But yes we can change tracking result to get them thanks

targos:
> But yes we can change tracking result to get them thanks

No need to change tracking result. The information are available in `result.event`

wadjih-bencheikh18:
> No need to change tracking result. The information are available in `result.event`

could you please explain how ? because i did't find how to get max / min

wadjih-bencheikh18:
and also we're not inside plotContext provider

targos:
> could you please explain how ? because i did't find how to get max / min

I suggested something in my review of your other pull request.

I'm going to close this PR for now because we think that keyboard events can always be handled from the parent of `Plot` as they don't need access to the plot's state.

PR #337, opened 2022-02-15
Feat : Fix legend

wadjih-bencheikh18:
- set position props default  value as embedded closes : https://github.com/zakodium/react-plot/issues/321
- fix legend showHide in barSeries
- remove hard coded value in legend ( we can now change fontSize of label  & StrokeWidth in lineStyle


targos:
show/hide doesn't work anymore in the stories

wadjih-bencheikh18:
the same problem was in https://github.com/zakodium/react-plot/pull/294
it works perfectly in local and i didn't change on how it works
@stropitek could please help here, you fixed the problem in the previous PR

targos:
ok, let's merge and see if it works with all the latest code

Issue #338, opened 2022-02-16
useLassoSelection


lpatiny:
This advanced hook would allow to select points in scatter series.

- this would useDrawPath
- by default it should be filled. I would suggest contour line in black 2 points thick and filled with black translucent: 0.8
- accepts a callback that receives for each scatter series (and only the scatter series) all the points that are inside the lasso selection
- we need a way to highlight the selected points that are selected

lpatiny:
Could first be done in a story, no need to create any hook.

The following dependency could be used: https://www.npmjs.com/package/point-in-polygon

We can take the PCA data of ecstasy but in a new story.

![2022-02-17 11 05 14](https://user-images.githubusercontent.com/1484241/154453087-fc7723f0-df07-4035-9afb-e9a2ff0fc2d2.gif)

wadjih-bencheikh18:
i see that useDrawPath is made using polyline and we should make a new version with polygone (add a prop if we want it polyline /polygone ) to be like the example (the first point is attached to the last one )

PR #339, opened 2022-02-17
Feat: UseLassoSelection hook

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/338


stropitek:
Can you add the callback that gets called when the user releases the pointer, which will be called with the selected points

You can log those values in a storybook action

targos:
In https://github.com/zakodium/react-plot/issues/338 we asked to only do a story (not a hook in src)

stropitek:
If the only difference between polygone and polyline is weather the returned shape is closed or not, then you can rename the `type` option to `close` options which is a boolean (`true` => Polygone, `false` => Polyline)

wadjih-bencheikh18:
> In #338 we asked to only do a story (not a hook in src)

yes in the first commit then i created just a story then it was easy to make it a hooks

wadjih-bencheikh18:
> Can you add the callback that gets called when the user releases the pointer, which will be called with the selected points
> 
> You can log those values in a storybook action

i was trying to impliment this and i had problems
to change style i'm using callback style for scatter points but in the hook i don't really have those points
so i tried to add onDraw prop to scatter Series so it excute when we draw markers , but the scatter series render many times

stropitek:
You can rename `onDraw` -> `onEnd`. `onDraw` is semantically too close to `onDrawing`

> but the scatter series render many times

Can you explain in more details? To me you just have to call the same algorithm that computes points within the polygon `onEnd`, and then pass it to the callback. The callback won't update the graph, it should only log those points in a story action.

wadjih-bencheikh18:
> Can you explain in more details?

onEnd i don't have  the selected points (hovered)
 have only polygone points
 I'm changing selected points style with a style callback `  (point)=>  inPolygone? hoverStyle : defaultStyle  `

targos:
> yes in the first commit then i created just a story then it was easy to make it a hooks

Yes, but it's not clear yet what API we want to expose publicly. You can keep the hook but please move the back to the story.

src/hooks/types.ts
```diff
@@ -14,6 +14,6 @@ export interface RectangleOptions {
   style?: CSSProperties;
 }
 export interface PathOptions {
-  color?: CSSProperties['stroke'];
+  color?: CSSProperties['stroke' | 'color'];
```
targos, line 17: What does this do?

wadjih-bencheikh18: i fixed it , it's suppoosed to be fill , because we have polygone or polyline

src/hooks/useDrawPath.tsx
```diff
@@ -1,21 +1,30 @@
...
 
-export interface UseDrawPathOptions extends DualAxisOptions, PathOptions {
-  onDraw?: (path: { data: Record<string, number>[] }) => void;
+export interface UseDrawPathOptions
+  extends ControllerHookOptions,
+    DualAxisOptions,
+    PathOptions {
+  close?: boolean;
+  onDraw?: (points: SeriesPoint[]) => void;
+  onEnd?: (points: SeriesPoint[]) => void;
```
targos, line 16: this is inconsistent with useDrawRectangle, which calls `onDraw` at the end, not on each move.

wadjih-bencheikh18: > You can rename `onDraw` -> `onEnd`. `onDraw` is semantically too close to `onDrawing`

it was 
onDraw for the onEnd 
and onDrawing for onMove

stropitek: I'm still in favor of `onEnd`, but indeed it would have to be used consistently in all hooks.

I don't like `onDraw` because I understand it as "when the hook draws something, call this callback", but the hook in itself actually stops drawing on end

wadjih-bencheikh18: i also agree with @stropitek

targos: it's good to me if we also rename the callback in the other hook.

stropitek (commented):
LGTM once the thing with the callback names is sorted out

wadjih-bencheikh18:
> LGTM once the thing with the callback names is sorted out

callbacks names fixed

PR #341, opened 2022-02-18
feat: interpret padding as number, string, or percentage

targos:
BREAKING CHANGE: Numeric padding is now interpreted as an offset on the axis' scale.
Use for example `paddingStart="10%"` instead of `paddingStart={0.1}` for percentages.

Closes: https://github.com/zakodium/react-plot/issues/319

### DIFF
--- a/stories/axis-boundaries.stories.tsx
+++ b/stories/axis-boundaries.stories.tsx
@@ -12,27 +12,12 @@ export default {
     xMax: 6,
     yMin: 0,
     yMax: 6,
-    paddingLeft: 0.01,
-    paddingRight: 0.01,
-    paddingTop: 0.01,
-    paddingBottom: 0.01,
   },
 } as Meta;
 
 type Props = Record<string, number>;
 export function Control(props: Props) {
-  const {
-    width,
-    height,
-    xMin,
-    xMax,
-    yMax,
-    yMin,
-    paddingLeft,
-    paddingRight,
-    paddingTop,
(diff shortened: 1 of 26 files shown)


stories/axis-boundaries.stories.tsx
```diff
@@ -12,10 +12,10 @@ export default {
     xMax: 6,
     yMin: 0,
     yMax: 6,
-    paddingLeft: 0.01,
-    paddingRight: 0.01,
-    paddingTop: 0.01,
-    paddingBottom: 0.01,
+    paddingLeft: '1%',
```
lpatiny, line 15: I could not let work this example 

http://localhost:6006/?path=/story/plot-axis-boundaries--control&args=paddingLeft:1

and in contrast with other examples like

http://localhost:6006/?path=/story/api-axis--axis-bottom

 we don't see in the storybook that the property may be either a number of a string.

lpatiny (commented):
The functionnalities are working as I expect. Great feature !

lpatiny:
It works as I would expect. Great new functionnality !

stories/axis-boundaries.stories.tsx
targos: Yeah, I will add that to the breaking changes too and remove the controls from this story. It's not possible to specify both paddings and min/max, as we discussed because they conflict with each other, especially when zoom interactions enter the game.

PR #344, opened 2022-02-19
feat!: use pointer events instead of mouse events

targos:
This allows to support touch screens and pen devices.


targos:
@lpatiny do you want to check a few stories with your iPad? I checked with my phone and touch screen laptop.

lpatiny:
Actually storybook is not working in safari (even on OsX)

image

I guess it is related to our nmr-processing library:

```
node_modules/nmr-processing/lib-esm/utilities/resurrectRange.js:        .split(/(?: |(?<=[0-9])-)/)
node_modules/nmr-processing/lib/utilities/resurrectRange.js:        .split(/(?: |(?<=[0-9])-)/)
node_modules/nmr-processing/src/utilities/resurrectRange.ts:    .split(/(?: |(?<=[0-9])-)/)
node_modules/v8-to-istanbul/lib/source.js:    for (const [i, lineStr] of source.split(/(?<=\r?\n)/u).entries()) {
```

I will check this

targos:
It's not the first time that we have this issue. I remember that nmrium was broken because of this.

https://caniuse.com/js-regexp-lookbehind

lpatiny:
In react-plot it is related to iv example

image

We should disable it for now I think.

targos:
iv example removed

lpatiny:
The storybook build is still not working in safari with the following error:

```
SyntaxError: Duplicate parameter 'height' not allowed in function with default parameter values.
```

But `npm run dev` works correctly in Safari on the other hand.

targos:
If we look at how they fixed a similar issue in https://github.com/jupyterlab/jupyterlab/pull/7993/files
It seems that we need to find a function that has the same name as one of its parameters. Do you have a more precise indication of where the error comes from?

lpatiny:
```
            },
            height: function height(height, box=this.bbox()) {
                return height == null
                ? box.height
                : this.size(box.width, height, box);
            },
 ```

**Comment by lpatiny** [comment]
It is in 'ContainerGeometry'

```
        let containerGeometry = {
            __proto__: null,
            dmove: function dmove(dx, dy) {
                return (
                this.children().forEach((child, i) => {
                    let bbox;
                    try {
                        bbox = child.bbox();
                    } catch (e) {
                        return;
                    }
                    const m = new Matrix(child);
                    const matrix = m.translate(dx, dy).transform(m.inverse());
                    const p = new Point(bbox.x, bbox.y).transform(matrix);
                    child.move(p.x, p.y);
                }),
                this
                ) ;
            },
            dx: function dx(dx) {
                return this.dmove(dx, 0);
            },
            dy: function dy(dy) {
                return this.dmove(0, dy);
            },
            height: function height(height, box=this.bbox()) {
                return height == null
                ? box.height
                : this.size(box.width, height, box);
            },
            move: function move(x=0, y=0, box=this.bbox()) {
                const dx = x - box.x;
                const dy = y - box.y;
                return this.dmove(dx, dy);
            },
            size: function size(width, height, box=this.bbox()) {
                const p = proportionalSize(this, width, height, box);
                const scaleX = p.width / box.width;
                const scaleY = p.height / box.height;
                return (
                this.children().forEach((child, i) => {
                    const o = new Point(box).transform(new Matrix(child).inverse());
                    child.scale(scaleX, scaleY, o.x, o.y);
                }),
                this
                ) ;
            },
            width: function width(width, box=this.bbox()) {
                return width == null ? box.width : this.size(width, box.height, box);
            },
            x: function x(x, box=this.bbox()) {
                return x == null ? box.x : this.move(x, box.y, box);
            },
            y: function y(y, box=this.bbox()) {
                return y == null ? box.y : this.move(box.x, y, box);
            },
        };
        class G extends Container {
            constructor(node, attrs=node)
            {
                super(nodeOrNew('g', node), attrs);
            }
        }
        extend(G, containerGeometry),
        ```

I'm afraid it is in node_modules/@svgdotjs/svg.js/dist/svg.node.js

I'm checking

targos:
It is a dependency of "mass-tools":

└─┬ mass-tools@0.60.31
  └─┬ ms-report@1.4.0
    └── @svgdotjs/svg.js@3.1.1

lpatiny:
The original problem seems to be there:

https://github.com/svgdotjs/svg.js/blob/2b028c35ab1a77c38a850c3c21082a5dacc18ac8/src/elements/A.js#L13

I'm checking not to use mass-tools

lpatiny:
It works correctly now on the iPad. Maybe we should add an issue to allow two fingers vertical scale ???

But locally I have now a build problem even after deleting node-modules and `npm i`

```
> react-plot@0.16.0 build-ts
> tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json

src/hooks/usePlotControllerAxis.ts:8:17 - error TS4058: Return type of exported function has or is using name 'PlotAxisOverrides' from external module "/Users/lpatiny/git/zakodium/react-plot/src/contexts/plotController/usePlotOverrides" but cannot be named.

8 export function usePlotControllerAxis(options: UsePlotControllerAxisOptions) {
                  ~~~~~~~~~~~~~~~~~~~~~


Found 1 error.
```

targos:
We can, but to support two fingers we need to listen to another kind of events (touch events)

PR #345, opened 2022-02-19
docs: add zoom to 2d spectrum

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/315


targos:
Let's ask @stropitek since he created the issue

stropitek:
No it's fine don't create the tracking. We will think about whether it makes sense to have a zoom when there is a shift in pixel. Because the zoom behavior here is really not intuitive.

PR #346, opened 2022-02-20
Feat : Gradient lineSerie

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/50


lpatiny:
The issue https://github.com/zakodium/react-plot/issues/50 is currently an open discussion and should not be implemented (nobody was assigned to this issue BTW). The colour is not a gradient but relies on experimental results and we should discuss in the issue how it could be implemented.

PR #349, opened 2022-02-20
docs: Weather Chart

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/79

Story: https://deploy-preview-349--react-plot.netlify.app/?path=/story/examples-weather-chart--weather-chart


lpatiny:
Please only put the bar series for the rain and  not the range series

Please fix the unit of the rain: (mm/h)

For the rain scale please put it between min=0, max=5

Please add some vertical grid (1px, light grey)

What are there those vertical white lines ? Is this a white grid ???

image

wadjih-bencheikh18:
no there is no vertical grid the data is made of diffrent lineseries / rangeSeries that are not perfectly linked

wadjih-bencheikh18:
> For the rain scale please put it between min=0, max=5

i added an issue for this

lpatiny:
> no there is no vertical grid the data is made of different lineseries / rangeSeries that are not perfectly linked (the last point in a Serie is not the start of the next one

Ok seems normal. Please prepare the data so that you have only one series rather than one series per day. This will solve the problem (and is the correct way in fact).

wadjih-bencheikh18:
@lpatiny fixed

lpatiny:
But I don't see the bar series with the rain anymore. Nothing is blue in the bottom

image

wadjih-bencheikh18:
> Nothing is blue in the bottom

its working perfectly in local 
image

the problem is in the draft deployment

wadjih-bencheikh18:
the reason why barseries is not working here is the same as mass example

lpatiny:
Ok so now that the other PR is merged: https://github.com/zakodium/react-plot/pull/351 it should be fixed if we rebase.

PR #351, opened 2022-02-20
docs : fix mass example story

wadjih-bencheikh18:


wadjih-bencheikh18:
@lpatiny it's fixed now 
all other series are tested too

PR #352, opened 2022-02-23
fix: axes domain

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/348


wadjih-bencheikh18:
@lpatiny do you want rainfall axis to have min=0 max=5 or calculate them auto

PR #358, opened 2022-02-26
fix: avoid feedback loop with shifted series

targos:
Closes: https://github.com/zakodium/react-plot/issues/357

For now, the bug is "fixed" by not applying the shift to the axis' min/max anymore.
I think we should discuss what is the expected behavior with a few simple cases and then implement the final min/max calculation inside the plot context, so the series cannot trigger infinite rerenders.

### DIFF
--- a/src/components/Series/RangeSeries.tsx
+++ b/src/components/Series/RangeSeries.tsx
@@ -38,7 +38,7 @@ export function RangeSeries<T extends RangeSeriesPoint>(
     yShift: propsYShift = '0',
   } = props;
 
-  const { xShift, xShiftInverted, yShift, yShiftInverted } = useShift({
+  const { xShift, yShift } = useShift({
     xAxis,
     yAxis,
     xShift: propsXShift,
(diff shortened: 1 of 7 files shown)


targos:
Here's a simple example where we can discuss expected behavior: https://deploy-preview-358--react-plot.netlify.app/?path=/story/examples-shift--single-series-example

targos:
First question: should the shift push min and max, or only extend one end?
For example [with xShift:1](https://deploy-preview-358--react-plot.netlify.app/?path=/story/examples-shift--single-series-example&args=xShift:1):
- push min and max -> min=0, max=5
- extend one end -> min=-1 (unchanged), max=5

wadjih-bencheikh18:
while calculating the new max/min after shift
if the shift is a number => theres not calculation using plot size => optimised 
the problem is when shift is a percentage or in px

wadjih-bencheikh18:
i tried spectrum using shift as a number
```
          xShift={i * 0.01}
          yShift={i * 10000000}
``` 
and it works

wadjih-bencheikh18:
i think https://github.com/zakodium/react-plot/issues/362 is due to the same problem
adding this in convertToPx will help optimising plots without shifts 
`if (scale === undefined || value === '0' || value === 0) return 0; `

wadjih-bencheikh18:
i searched in old PR 
i found that this story is crushing before adding shifts
release 0.16.0  https://github.com/zakodium/react-plot/pull/313
https://deploy-preview-313--react-plot.netlify.app/?path=/story/experimental-spectra-nmr--stack-spectra

wadjih-bencheikh18:
but we  still can see the diffrence between 
our version now :
https://zakodium.github.io/react-plot/?path=/story/experimental-spectra-nmr--stack-spectra
before shift : https://github.com/zakodium/react-plot/pull/311
https://deploy-preview-311--react-plot.netlify.app/

PR #359, opened 2022-02-26
docs: optimise spectrum2d

wadjih-bencheikh18:


targos:
I already did this optimization in https://github.com/zakodium/react-plot/pull/358

wadjih-bencheikh18:
> I already did this optimization in #358

yes im just trying to see the result

PR #361, opened 2022-02-26
chore!: displayMarker to displayMarkers

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/355


targos:
@wadjih-bencheikh18 I think `refactor` is better than `chore`. Also, the verb is missing at the beginning of the message. I changed it to:
```
refactor!: rename `displayMarker` prop to `displayMarkers`
```

PR #363, opened 2022-03-06
fix: nmr story

wadjih-bencheikh18:
closes : https://github.com/zakodium/react-plot/issues/362

### DIFF
--- a/src/components/Series/ScatterSeries.tsx
+++ b/src/components/Series/ScatterSeries.tsx
@@ -58,8 +58,8 @@ export function ScatterSeries<T extends SeriesPoint = SeriesPoint>(
     label,
     hidden,
     displayErrorBars = false,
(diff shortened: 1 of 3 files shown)


wadjih-bencheikh18:
@targos i think optmise data is the only solution to make it render fast (d3 line generator take time to calculate)
as mentioned here https://github.com/zakodium/react-plot/issues/316

wadjih-bencheikh18:
in https://blog.scottlogic.com/2020/05/01/rendering-one-million-points-with-d3.html 
they mention that : "SVG charts can handle around **1,000 datapoints**"

i'm not sure if there is a way to optmise scripting here

targos:
Yeah I agree this has to be fixed by reducing the number of points in the SVG (https://github.com/zakodium/react-plot/issues/316)

Issue #379, opened 2022-04-03
FunctionSerie start from 0 bug


wadjih-bencheikh18:
in the next example xAxis must start from 4

<Plot width={800} height={300}>
      <FunctionSeries getY={(x) => 5 * Math.sin(x)} />
      <LineSeries
        data={[
          { x: 4, y: 2 },
          { x: 5, y: -3 },
          { x: 6, y: 4 },
          { x: 7, y: 2 },
          { x: 8, y: 5 },
        ]}
      />
    </Plot>
    
    
![image](https://user-images.githubusercontent.com/56769876/161434504-a7a1f3cb-ad15-42b3-a8df-a48a45b9cb26.png)

Issue #378, opened 2022-04-03
Limite Markers & PointLabel in FunctionSerie


wadjih-bencheikh18:
we must add step prop
when we displayMarkers
![image](https://user-images.githubusercontent.com/56769876/161433785-70f5541b-7653-4bdb-bd54-f812fe3da57b.png)

when we displayPointLabels
![image](https://user-images.githubusercontent.com/56769876/161433812-0ffe415f-2589-4bcd-8911-2a2e3b4c1c00.png)

lpatiny:
Please do not allow those props in function series

Issue #377, opened 2022-04-03
Remove ErrorBar from FunctionSerie


wadjih-bencheikh18:

lpatiny:
Please do not allow this prop in FunctionSerie (and other props that do no make sense)

Issue #316, opened 2022-02-07
react-plot and series containing 500000 points


lpatiny:
React plot zoom slows down when there are too many points.

At least in http://react-plot.zakodium.com/experimental/nmr (spectrum in the bottom).

Here is an example of file:

[nmr13c.json.zip](https://github.com/zakodium/react-plot/files/8017330/nmr13c.json.zip)

This file contains 520000 points. It is in the form {x:[], y:[]} and can be converted to {x,y}[] using https://github.com/mljs/spectra-processing/blob/master/src/xy/xyToXYObject.ts

I would propose to use the following code in order to reduce the number of points if required. This has shown very good results in NMRium:

https://github.com/aiday-mar/spectra-processing/blob/0572f5c047ab23ac233a482ff00127c09c811a09/src/xy/xyReduce.ts#L9-L18

This function requires {x:[], y:[]} and you can always convert it using:

https://github.com/mljs/spectra-processing/blob/master/src/xyObject/xyObjectToXY.ts

Because the original data may contain extra information (other properties than x and y) this reduced version should only be used for SVG and not for tracking for example.

@stropitek Seems to me that visually there is no difference with this reduced version of the data and I think this could be the default behaviour before trying to generate the SVG we should try to reduce the number of points based on the current display width and the current from / to zoom factor (use a Memo for this and a local state ???).

lpatiny:
@stropitek just edit this issue so that it makes more sense at the implementation point of view if we are going in this direction.

stropitek:
It seems like the method to reduce the points is preserving the original points which is good. @lpatiny can you confirm it's the case?

However we cannot use this method as is because it returns the points and not the indices of the points that were preserved. Therefore any additional data linked to a point (for example error bar), or other data we potentially want to show to the user is lost.

lpatiny:
This reduction of number of points should only be applied if xIsMonotone https://github.com/mljs/spectra-processing/blob/master/src/x/xIsMonotone.ts

lpatiny:
We would rather add a new property that could reduce the number of points if the user wants.

onBeforePlot( points, from, to ) => points

targos:
Maybe we should reconsider this for later, after we implement the multi-svg layer optimizations?

lpatiny:
LineSeries could accept a callback as data but in this case it would requires boundary.

```
[...]

PR #380, opened 2022-04-08
fix: FunctionSeries

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-plot/issues/377
closes :  https://github.com/zakodium-oss/react-plot/issues/378
fixes : https://github.com/zakodium-oss/react-plot/issues/379
closes : https://github.com/zakodium-oss/react-plot/issues/385


wadjih-bencheikh18:
The problem now with `FunctionSeries` after these changes is we have to specifie max,min of vertical axis too

**Implementation detail:**
how FunctionSeries work?
it take max/ min from horizantal axis if this latter exist and draw lineSeries in [min, max] domain

to specifie max / min we need to either:
- axis max, min props
- or add another series => axis will calculate automatically max, min

=> logically we need just to specifie max /min of horizontal axis to draw it .
but the problem is When Plot is empty 
```
<Plot {...DEFAULT_PLOT_CONFIG}>
      <FunctionSeries
        getY={getY}
      />
      <Axis
        min={0}
        max={20}
        position="bottom"
      />
      <Axis position="left"/>
</Plot>
```
 => vertical axis domain is not defined => no horizantal axis

image

targos:
I hope the above will also be fixed with https://github.com/zakodium-oss/react-plot/issues/316

targos:
The new story updates infinitely and crashes. Is that expected?

wadjih-bencheikh18:
here i made simple solution but not the best ,
i just added ~ in functionSeries id so it won't be considered while calculating Max, Min

Issue #376, opened 2022-04-03
ErrorBars not working in BarSeries


wadjih-bencheikh18:
when displayMarker=false (default) we cannot display ErrorBars
ErrorBars should be independent of markers

Issue #374, opened 2022-04-03
Point Label doesn't work in LineSeries & barSeries


wadjih-bencheikh18:
when displayMarker=false (default) we cannot display point label
I think point label should be independent of markers

PR #381, opened 2022-04-08
fix: ErrorBar & PointLabel

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-plot/issues/376
closes : https://github.com/zakodium-oss/react-plot/issues/374


targos:
Something is wrong. See https://deploy-preview-381--react-plot.netlify.app/?path=/story/api-annotations--annotation-arrow and other stories which now has markers.

wadjih-bencheikh18:
> 

displayMarkers default value fixed

Issue #371, opened 2022-03-31
ScatterSeries legend's markershape


wadjih-bencheikh18:
https://github.com/zakodium-oss/react-plot/blob/60d1782dd813868a4674ee29ad963842b5477817/src/components/Series/ScatterSeries.tsx#L76-L88

there is a possibilty to have many markershapes in one scatterseries so it's hard to choose one

stropitek:
In general if the legend is complicated to generate automatically, the user should opt out of automatic legend and generate a manual legend IMO

wadjih-bencheikh18:
yes but the problem is when it's not complicated , makershape="star" we still have a circle in the legend
![image](https://user-images.githubusercontent.com/56769876/161287115-bd4d504c-7fb6-4b18-ab95-fb25b87672e4.png)

lpatiny:
Please use for now the style of the first point.

PR #383, opened 2022-04-08
Fix ScatterSeries Simple markersShape in Legend

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-plot/issues/371


src/components/Series/ScatterSeries.tsx
```diff
@@ -97,6 +100,7 @@ export function ScatterSeries<T extends SeriesPoint = SeriesPoint>(
     legendDispatch,
     otherProps.markerShape,
     otherProps.markerStyle?.fill,
+    props.markerShape,
```
targos, line 103: `otherProps.markerShape` is already here and contains the same value.

wadjih-bencheikh18: fixed

Issue #387, opened 2022-04-08
Infinite loop in every Plot


wadjih-bencheikh18:
i found that there's an infinite loop in every plot 

after searching i found that the problem is in this useMemo 
https://github.com/zakodium-oss/react-plot/blob/ae14c2c36260f258e63a95eaad93b38190fe1611/src/contexts/plotContext.ts#L192

after checking hook dependency, the problem was with `state.axes`

when i removed `state.axes` dependency from the hook everything is running perfectly (even NMR example)

PR #388, opened 2022-04-08
Optimisation

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-plot/issues/387

### DIFF
--- a/stories/spectra/nmr.stories.tsx
+++ b/stories/spectra/nmr.stories.tsx
@@ -49,6 +49,7 @@ export function StackSpectra() {
       xAxis="x"
       yAxis="y"
       lineStyle={{ stroke: color }}
+      yShift={`${i * -7}`}
     />
   ));
   return (
--- a/stories/spectra/nmr.stories.tsx
+++ b/stories/spectra/nmr.stories.tsx
@@ -61,8 +61,7 @@ export function StackSpectra() {
         position="left"
         label="Intensity / arbitrary"
         hidden
-        paddingStart="10%"
-        paddingEnd="10%"
+        paddingEnd="50%"
       />
     </Plot>
   );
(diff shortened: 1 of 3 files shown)


targos:
I don't think that can be the right fix. If this value is not in the dependencies, we could depend on an outdated version of it.
If the problem is that `state.axes` changes too much, the solution should be to prevent that from happening.

wadjih-bencheikh18:
> The solution should be to prevent that from happening.

No problem, going to search for a solution

wadjih-bencheikh18:
@targos 
i found the problem and fix it

wadjih-bencheikh18:
while fixing the problem i found that in time Axis , `tickLabelFormat` is not specified in AxisContext so after merging this PR .
i will fix it

stories/spectra/nmr.stories.tsx
```diff
@@ -49,7 +49,6 @@ export function StackSpectra() {
       xAxis="x"
       yAxis="y"
       lineStyle={{ stroke: color }}
-      yShift={`${i * -7}`}
```
targos, line 52: can we keep the shift and padding as before?

wadjih-bencheikh18: Yes no probleme, 
i just thought that the correct graph is without shift

wadjih-bencheikh18: fixed

stories/spectra/nmr.stories.tsx
```diff
@@ -61,7 +61,8 @@ export function StackSpectra() {
         position="left"
         label="Intensity / arbitrary"
         hidden
-        paddingEnd="50%"
+        paddingStart="10%"
+        paddingEnd="10%"
```
targos, line 65: please also keep the original padding

wadjih-bencheikh18: fixed

PR #396, opened 2022-04-21
feat: Secondary grids

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-plot/issues/384


src/components/Axis/HorizontalAxisGridLines.tsx
```diff
@@ -1,35 +1,82 @@
...
+        style={style}
+      />
+    ));
+
+    if (secondaryGrid) {
+      const density = 5;
+      const secondaryTicks = scale?.ticks(primaryTicks.length * density) || [];
+      secondaryTicks.forEach((tick) => {
+        const position = scale?.(tick);
+        if (
```
targos, line 49: don't we already compute the secondary ticks somewhere (to display them)? We could move that higher in the tree to avoid computing them in multiple places.

wadjih-bencheikh18: this is not secondaryTicks, just need to rename the variable
also secondaryTicks and secondaryGrid are not the same always

wadjih-bencheikh18: we can add `secondaryTicksDensity` , `secondaryGridDensity` props in `Axis`

targos (approved):
Thanks!

PR #399, opened 2022-04-28
feat: sanplot example with big data

hamed-musallam:


targos (changes requested):
Please move it to the examples folder, and make sure it appears in the Examples category in the storybook interface.

Issue #401, opened 2022-04-28
Display secondary grid lines independently of primary


targos:
Currently, `displaySecondaryGridLines` only works if `displayPrimaryGridLines` is also set.
@wadjih-bencheikh18  Would it be difficult to make both options independent?

wadjih-bencheikh18:
> Would it be difficult to make both options independent?

No i think it's simple

PR #403, opened 2022-04-28
feat: make primary & secondary grids independent

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-plot/issues/401


wadjih-bencheikh18:
> Can you take care of the `useMemo` warnings?

fixed

Issue #406, opened 2022-05-13
Create a box-plot annotation


lpatiny:
![image](https://user-images.githubusercontent.com/1484241/168286048-be0f5580-f2db-4b84-959a-2b691dd871ec.png)

q1
median
q3
min
max

lpatiny:
![image](https://user-images.githubusercontent.com/1484241/168292117-4b051ca6-558b-47ba-94f8-90d70ecca21b.png)

We have identified couple of properties that could be used. Some of them are in our 'units' system, pixel, axis value or percent:
- width (the width of the box plot on the yAxis
- y the position on the yAxis
- min, max, q1, q3 and median, the values on the xAxis

We also identified various properties to change the style of the box plot

Please take care  that we can invert xAxis and yAxis !!! This means that if we want to draw a vertical box plot we should 'simply' add
- xAxis: 'y'
- yAxis: 'x'

Default values could be no fill, black color, 1 px width except for median that would be 2 px wide.

![image](https://user-images.githubusercontent.com/1484241/168294539-4d94b9a9-6a8c-4909-ba1e-e7ca9cbb837e.png)

@wadjih-bencheikh18 could you try to implement this and of course tell us if we forgot something or if it seems weird.

One of the example in storybook could be something like

![image](https://user-images.githubusercontent.com/1484241/168296003-3d1a26d1-fdb3-4bc8-bb66-5ab4019ff829.png)

In which we display horizontal axis and not the vertical axis. But we should also have an horizontal axis from 0 to 1000 but the box plot that has the values 200,300,400,500,700.

stropitek:
> * width (the width of the box plot on the yAxis

This seems confusing to me since width is usually related to the horizontal axis. `minMaxWidth` being an horizontal property but `width` being a vertical property seems confusing to me.

Boxplots are most often vertical from what I've seen in publications so maybe that should be the behavior with `xAxis=x` and `yAxis=y`. Then we could keep `width` but `minMaxWidth` -> `minMaxHeight`, etc...

stropitek:
Other possibility might be to rename "width" to "size" or "breadth"? I personally prefer "breadth" to "width"

targos:
Not sure about the main width/breadth, but I think we should keep minMaxWidth, as it refers to the stroke width, not a particular dimension

wadjih-bencheikh18:
> minMaxWidth

[...]

PR #408, opened 2022-05-13
Box plot

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-plot/issues/406


src/components/Annotations/BoxPlot.tsx
```diff
@@ -0,0 +1,101 @@
...
+    medianWidth,
+    medianColor = 'black',
+    strokeWidth,
+    strokeColor = 'black',
+    fillColor = 'none',
+    whiskerWidth,
+    whiskerColor = 'black',
+    minMaxWidth,
+    minMaxColor = 'black',
+    ...positon
```
targos, line 36: typo: position

stories/annotations/boxPlot.stories.tsx
```diff
@@ -0,0 +1,32 @@
+import { Meta } from '@storybook/react';
+
+import {
+  AnnotationBoxPlotProps,
+  BoxPlot,
+} from '../../src/components/Annotations/BoxPlot';
```
targos, line 6: This should be imported from `'../../src'`. Need to export in https://github.com/zakodium-oss/react-plot/blob/main/src/components/Annotations/Annotation.tsx

wadjih-bencheikh18: to be imported from ../../src we must use annotation.BoxPlot

wadjih-bencheikh18: fixed in all annotations stories

targos: Cool, thanks.
I think it's better to use the public API in the storybook so we don't forget to export and it's easier to copy/paste an example.

src/components/Annotations/BoxPlot.tsx
```diff
@@ -0,0 +1,117 @@
...
+    <g>
+      {/* interquartile range */}
+      <rect
+        x={horizontal ? q1 : y1}
+        y={horizontal ? y1 : q1}
+        width={horizontal ? height : width}
+        height={horizontal ? width : height}
+        strokeWidth={strokeWidth}
+        stroke={strokeColor}
+        fill={boxColor}
```
targos, line 63: boxColor should be passed to stroke (black by default)

wadjih-bencheikh18: Luc mentioned in the issue that the default values of the box is
No fill and black stroke

targos: Yes that's the point.

if we do `stroke={boxColor}`, we will have a black stroke and no fill.

targos: This depends on the previous suggestion which is to remove `strokeColor` and `strokeWidth` from the props.

targos: If we want to fill the box, we can then do `boxStyle={{fill: 'red'}}`

wadjih-bencheikh18: Yes clear

src/components/Annotations/BoxPlot.tsx
```diff
@@ -0,0 +1,117 @@
...
+  strokeWidth?: number | `${number}`;
+  strokeColor?: string;
+  boxColor?: string;
+  boxStyle?: CSSProperties;
+  whiskerWidth?: number | `${number}`;
+  whiskerColor?: string;
+  whiskerStyle?: CSSProperties;
+  minMaxWidth?: number | `${number}`;
+  minMaxColor?: string;
+  minMaxStyle?: CSSProperties;
```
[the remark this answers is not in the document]
wadjih-bencheikh18: fixed

targos:
- Can you change the default values in the story so that there is no overlap between the spectrum and the boxplot?
- Also, please create another story with a vertical boxplot

wadjih-bencheikh18:
> * Can you change the default values in the story so that there is no overlap between the spectrum and the boxplot?
> * Also, please create another story with a vertical boxplot

done

but while creating vertical BoxPlot i was getting the next error :  `The axis x should be  top bottom and y should be left right` because validateAxis hook
so i added an onlyOrthogonal to make the hook test just that axis are orthogonal without take in consideration that x must be horizontal ...

stropitek:
image

- minMax should be above whiskers
- Box should be above whiskers
- Median should not extend past the box's border

wadjih-bencheikh18:
@stropitek
thank you , it's fixed now

stropitek (approved):
LGTM

Issue #412, opened 2022-05-23
BoxPlot q1 and q3 mixed up? ( when viewing vertically)


id2359:
Hello 
It seems BoxPlot is not showing the quartiles correctly when BoxPlot is vertical ( I think) I was adding a demo to the react-plot-docs repo and noticed the boxplot only displayed correctly if I swapped q1 and q3.

I used ml-stats to get quartiles q1 and q3 of some test data, but the BoxPlot only appears correctly when q3 is supplied to the q1 property ( and q1 to q3) :

viz:

```
import { median, quartiles, max, min } from 'ml-stat/array';

const quarts = quartiles(data);
// these are correctly calculated by ml-stat 
//  q1 is the lower quartile , q1 < q3 
q1 = quarts.q1; 
q3 = quarts.q3; 
But in constructing the BoxPlot I had to do:
 const props: AnnotationBoxPlotProps = {
    max: max(data),
    min: min(data),
    q1: quarts.q3,
    median: median(data),
    q3: quarts.q1,
    width: 1.0,
    y: 0,
    xAxis: 'y',
    yAxis: 'x',
    medianColor: 'orange',
    medianStyle: { strokeWidth: 4 },
    minMaxStyle: { strokeWidth: 2 },
    minMaxColor: 'blue'
  };
```

targos:
Thanks for the report.

PR #413, opened 2022-05-23
fix: vertical boxPlot

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-plot/issues/412


stories/annotations/boxPlot.stories.tsx
```diff
@@ -49,8 +49,8 @@ Vertical.args = {
   xAxis: 'y',
   yAxis: 'x',
   y: 2500,
-  q1: 80,
-  q3: 50,
+  q1: 50,
+  q3: 80,
   width: 300,
   min: 90,
   max: 20,
```
targos, line 56: Can you also invert these?

Issue #417, opened 2022-06-23
disable hooks option


wadjih-bencheikh18:
in order to implement https://github.com/zakodium-oss/analysis-ui-components/issues/144

PR #419, opened 2022-06-23
Disable hooks

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-plot/issues/417

### DIFF
--- a/src/contexts/plotController/plotControllerContext.tsx
+++ b/src/contexts/plotController/plotControllerContext.tsx
@@ -46,11 +46,12 @@ const plotEventsUserContext =
 
 export function usePlotEvents(
   handlers: EventsHandlers,
-  options?: ControllerHookOptions,
+  options: ControllerHookOptions = {},
 ) {
-  const id = options?.controllerId;
-  const userContext = plotEventsUserContext.useNestedContext(id);
-  if (!userContext) {
+  const { controllerId: id, disabled = false } = options;
+
+  const userContext = !disabled && plotEventsUserContext.useNestedContext(id);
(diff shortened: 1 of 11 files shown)


src/contexts/plotController/plotControllerContext.tsx
```diff
@@ -46,11 +46,12 @@ const plotEventsUserContext =
...
   handlers: EventsHandlers,
-  options?: ControllerHookOptions,
+  options: ControllerHookOptions = {},
 ) {
-  const id = options?.controllerId;
-  const userContext = plotEventsUserContext.useNestedContext(id);
-  if (!userContext) {
+  const { controllerId: id, disabled = false } = options;
+
+  const userContext = !disabled && plotEventsUserContext.useNestedContext(id);
```
targos, line 53: It is not allowed to call a hook conditionally. We need to always call `useNestedContext`

src/contexts/plotController/plotControllerContext.tsx
```diff
@@ -60,11 +61,15 @@ export function usePlotEvents(
...
+    }
+  }, [disabled, handlers]);
   useEffect(() => {
-    userContext.registerHandlers(ref);
-    return () => userContext.unregisterHandlers(ref);
+    if (userContext) {
+      userContext.registerHandlers(ref);
+      return () => userContext.unregisterHandlers(ref);
+    }
   }, [userContext]);
```
targos, line 73: I'm not sure it will work correctly if `disabled` changes dynamically. Can you add a story with a button to toggle the value of `disabled` ?

Issue #423, opened 2022-06-24
Bug


wadjih-bencheikh18:
![image](https://user-images.githubusercontent.com/56769876/175521351-41587e31-340f-49b1-9e29-430f112ee560.png)

PR #424, opened 2022-06-24
fix: no default axis hooks bug

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-plot/issues/423


targos:
Do we already have a story with this case? If not, can you add one?

wadjih-bencheikh18:
> Do we already have a story with this case? If not, can you add one?

the problem is in https://github.com/zakodium-oss/analysis-ui-components/pull/166

https://deploy-preview-166--analysis-ui-components.netlify.app/storybook/?path=/story/layout-measurementexplorer--control choose xVariable : "t"

wadjih-bencheikh18:
i will try to create a story with the problem

wadjih-bencheikh18:
@targos story added

targos:
How can I reproduce the problem with the story? I tried to add a log when the value is `undefined` but nothing happens.

targos:
I think it would be better to show the bug with a very simple story instead of this one.

wadjih-bencheikh18:
> I think it would be better to show the bug with a very simple story instead of this one.

the problem happend while changing the data after the plot loaded so i added a story where u can change horizantal and vertical variables

wadjih-bencheikh18:
the bug always happend if we revert this modification https://github.com/zakodium-oss/react-plot/pull/424/commits/2aac5826cb1cd4373d0c2cfe1eafccc6709185b4

targos:
But for me this is not the correct fix. The crosshair doesn't work as expected.

wadjih-bencheikh18:
@targos i found the problem and fix it 
the thing that hooks were not genralized to use with no default axis variables

also i added DualAxisOptions in useAxisZoom to fix the problem

targos:
Vertical zoom doesn't work (it's horizontal), but apart from that, good job! 
https://deploy-preview-424--react-plot.netlify.app/?path=/story/examples-measurement--measurement&args=xAxis:a;yAxis:t;hook:Vertical+Zoom

wadjih-bencheikh18:
> Vertical zoom doesn't work (it's horizontal)

Sorry about that , it's fixed now

Issue #428, opened 2022-07-18
legend bug


wadjih-bencheikh18:
when using function series under bar series the legend of this latter does not appear
```
 <Plot {...DEFAULT_PLOT_CONFIG}>
      <Legend {...props} labelStyle={{ cursor: 'hand' }} />
      <BarSeries
        data={data2}
        lineStyle={{
          stroke: 'green',
        }}
        label="Label Bar series"
      />
      <FunctionSeries
        getY={(x) => 5 * Math.sin(x) + 10}
        lineStyle={{
          stroke: 'brown',
        }}
        label="Label Function series"
      />
    </Plot>
```
![image](https://user-images.githubusercontent.com/56769876/179599172-e30c454a-6661-4ac4-a21e-f3d7077aefab.png)

but if we reverse it works correctly

PR #429, opened 2022-07-18
docs: add bug story

wadjih-bencheikh18:
closes : https://github.com/zakodium-oss/react-plot/issues/428

### DIFF
--- a/stories/control/legend.stories.tsx
+++ b/stories/control/legend.stories.tsx
@@ -4,6 +4,7 @@ import { useState } from 'react';
 import {
   Axis,
   BarSeries,
+  FunctionSeries,
   Legend,
   LegendProps,
   LineSeries,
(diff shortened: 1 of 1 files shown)


src/components/Series/BarSeries.tsx
```diff
@@ -60,18 +60,20 @@ export function BarSeries(props: BarSeriesProps) {
...
+        payload: {
+          id,
+          label: otherProps.label,
+          colorLine,
+          shape,
+        },
+      });
+    }
     return () =>
       legendDispatch({ type: 'REMOVE_LEGEND_LABEL', payload: { id } });
```
targos, line 75: the cleanup callback should also be in the `if` block (same in other files)

wadjih-bencheikh18: thank you, the bug is fixed now

Issue #457, opened 2022-12-28
Using react-plot to generate server side SVG


lpatiny:
We would need to test and provide examples (documentation) that allows to create SVG using react-plot on the server (without graphical interface).

targos:
https://beta.reactjs.org/reference/react-dom/server/renderToStaticMarkup

PR #458, opened 2022-12-28
test: react-plot on server side

wadjih-bencheikh18:
closes https://github.com/zakodium-oss/react-plot/issues/457


lpatiny:
I would like to be able to achieve something like:

```
import {createSVG} from 'react-plot';

const jsonData = {};

const svg = createSVG(jsonData, {width: 600, height: 400})

writeFileSync('test.svg', svg, 'utf8')
```

But we don't have a 'jsonData' format in which we could specify most of the properties from react-plot. @targos Do we want to go in this direction ?

PR #459, opened 2023-01-03
feat: implement createSVG to render react-plot svg in server-side

wadjih-bencheikh18:


targos:
I don't think we should expose such a method. It is easy enough and more flexible to directly call the React APIs where we need them server-side.

targos:
@lpatiny we discussed this and I thought you agreed?

